PR #117, opened 2026-04-19
Rust refactors

Random-Scientist:
* adds the `.vscode` user configuration folder to `.gitignore`
* updates `sable_rapier` and `marten` to the Rust `2024` edition (biggest change being that the `no_mangle` attribute is now properly marked `unsafe`).
* moves the two isolated independent projects into a single cargo workspace, which specifies shared dependencies once, as well as deduplicating them during development.
* replaces a bunch of low-hanging java-isms with cleaner and/or more idiomatic Rust counterparts.


common/build.gradle
```diff
@@ -120,7 +120,7 @@ tasks.register('copyRustNatives', Copy) {
     ]
 
     moves.forEach { map ->
-        from(file("$rustProjectDir/${map.src}")) {
+        from(file("$rustRootDir/${map.src}")) {
```
Random-Scientist, line 123: thanks, i wasn't sure if i needed to change the gradle script but apparently the answer is yes.

Ocelot5836: Yup! I tested the build script and made sure it works properly

ryanhcode:
all really good and well needed changes!

PR #138, opened 2026-04-19
Fix the command output issue for /sable physics rotation

cewno:
Prior to the fix, both 'set' and 'add' commands produced the output 'added'.

After the fix, it is functioning normally.


Ocelot5836:
Can you enable contributors can edit PRs?

cewno:
ok

cewno:
> Can you enable contributors can edit PRs?

I have enabled it. Now you can make commits to my branch.

PR #158, opened 2026-04-19
Add Russian localization for Sable commands

luffypirat:


common/src/main/resources/assets/sable/lang/ru_ru.json
```diff
@@ -0,0 +1,121 @@
+{
+  "menu.savingSubLevels": "Сохранение подуровней",
```
VladisCrafter, line 2: The direct translation of "sublevels" - "подуровни" - genuinely sounds a little awkward, an option I'd like you to consider is "физические конструкции" - close to how Simulated Contraptions are currently translated for Simulated/Aeronautics. It shouldn't cause much misunderstanding to the end user, even rather sounds quite intuitive.

Modog500000: I don't know anything about Russian, but in Sable the term is "sublevels" which should be translated to an equivalent that makes sense. However in simulated we call sublevels "simulated contraptions." The translation for these should be different in sable as opposed to simulated.

VladisCrafter: @Modog500000  let me explain my choice.

> but in Sable the term is "sublevels" which should be translated to an equivalent that makes sense.

First up, that's the problem. A literally translated term "sublevels" - "подуровни" - would make totally no sense to anyone who is not into programming or Minecraft's technical aspect. I do know what "level" is supposed to mean because of those `assert level != null`, but a regular user couldn't say the same. And genuinely outside of MC modding nobody calls worlds "levels" (at least in Russian).
To add on the Russian-specific part, "подуровни" is a) quite difficult to pronounce the way that the "уровни" ("levels") part is accented at, and b) to me it sounds a lot like "дурень", which is... "a bubblehead"...

> The translation for these should be different in sable as opposed to simulated.

Secondly, it is different. The term I suggest can roughly be as "physical structures", as opposed to "physical contraptions" in Simulated. If "why physical in the first place" is the question that pops up, it's quite simple: translating "simulated" as-is would imply engaging a neologism (which, in this case, is also a kinda "anglicism"), and neologisms generally don't fit within the steampunk aesthetic of Create nor the semi-medieval aesthetic of Minecraft. Maybe "simulated" has deeper roots in English language (or maybe you just don't care that much), but in Russian we couldn't just use it and call it a day.

VladisCrafter: @luffypirat with all that saying, I strongly recommend switching from a literal "подуровни" to at least a more understandable alternative. Like "Подмиры", as a last resort :P

luffypirat: > @luffypirat with all that saying, I strongly recommend switching from a literal "подуровни" to at least a more understandable alternative. Like "Подмиры", as a last resort :P

Это библиотека для модов и перевод технических команд, я не буду упрощать эти команды, ими будут пользоваться только прошареннын игроки и тестеры модов

Modog500000: > @Modog500000  let me explain my choice.
> 
> > but in Sable the term is "sublevels" which should be translated to an equivalent that makes sense.
> 
> First up, that's the problem. A literally translated term "sublevels" - "подуровни" - would make totally no sense to anyone who is not into programming or Minecraft's technical aspect. I do know what "level" is supposed to mean because of those `assert level != null`, but a regular user couldn't say the same. And genuinely outside of MC modding nobody calls worlds "levels" (at least in Russian).
> To add on the Russian-specific part, "подуровни" is a) quite difficult to pronounce the way that the "уровни" ("levels") part is accented at, and b) to me it sounds a lot like "дурень", which is... "a bubblehead"...
> 
> > The translation for these should be different in sable as opposed to simulated.
> 
> Secondly, it is different. The term I suggest can roughly be as "physical structures", as opposed to "physical contraptions" in Simulated. If "why physical in the first place" is the question that pops up, it's quite simple: translating "simulated" as-is would imply engaging a neologism (which, in this case, is also a kinda "anglicism"), and neologisms generally don't fit within the steampunk aesthetic of Create nor the semi-medieval aesthetic of Minecraft. Maybe "simulated" has deeper roots in English language (or maybe you just don't care that much), but in Russian we couldn't just use it and call it a day.

That's all fair, and I am not saying that your wrong (again I don't know Russian lol) just was clarifying that sublevels and simulated contraptions are NOT the same thing. Also side note but "level" isn't used in English to mean world either. It's just used in video games/computing, and it is usually used to represent disconnected sets of information to organize them.

VladisCrafter: > Это библиотека для модов и перевод технических команд, я не буду упрощать эти команды, ими будут пользоваться только прошареннын игроки и тестеры модов

Welp.

common/src/main/resources/assets/sable/lang/ru_ru.json
```diff
@@ -0,0 +1,121 @@
...
+  "camera_type.sub_level_view": "Камера подуровня активирована",
+  "camera_type.sub_level_view_unlocked": "Свободная камера подуровня активирована",
+  "force_group.sable.gravity": "Гравитация",
+  "force_group.sable.drag": "Лобовое сопротивление",
+  "force_group.sable.levitation": "Левитация",
+  "force_group.sable.balloon_lift": "Подъёмная сила оболочки",
+  "force_group.sable.propulsion": "Тяга",
+  "force_group.sable.lift": "Подъёмная сила",
+  "force_group.sable.magnetic_force": "Магнитная сила",
+  "schematic.sable.mirror_not_supported": "Нельзя зеркалировать схемы, содержащие подуровни!",
```
justlofe, line 117: Maybe "отзеркаливать", not "зеркалировать"?
Да, зеркалировать чаще используется в техническом контексте, но в разговорном языке это слово практически не возникает, поэтому отзеркаливать выглядит уместнее.

PR #161, opened 2026-04-19
more aggressive rust refactoring

Random-Scientist:
this codebase will drive me insane but i'm too stubborn to go to sleep
the great de-nalgebrification.
* Use glam{x} {D, I}Vec3 as the canonical vector types
* as a consequence of the above, collapse a huge amount of manual implementations of vector arithmetic to vector methods/overloads
* pass vectors internally as vectors instead of constantly un and repacking them (rust won't punish you for wrapping your arguments in nice types)


Random-Scientist:
before marking this as ready for review i'll go over the diff myself and make sure i got all the operator precedences right 💀

Random-Scientist:
yep... got hit by an operator precedence bug in `get_overlapping_nodes`. I think that's the only one (at least, everything seems to work now when i mess around ingame, and i went over the changes pretty carefully in my second pass).

tazercopter:
sounds like I need to look at this for https://github.com/ryanhcode/sable/pull/125

ryanhcode:
we've added a CLA now btw, the workflow should update on the next commit

CherrieTheShifter:
Yes This

Random-Scientist:
ill revisit this over the new changes at some point

PR #162, opened 2026-04-19
add CBC compat

Tribulla:
add CBC compat (not sure what else to write here but if you have any questions ping me in aeronautics dev chat so I notice it)


ryanhcode:
We aren't adding CBC compat on our end, I believe the CBC developers are working on it on their end

Tribulla:
ok, are they though since VS has the compat on their side and it doesnt make sense to add the compat on CBC side cus I am not even sure it is possible? I could be wrong though.

Tribulla:
ok, I checked and they are indeed doing it on their side so I was wrong.

PR #176, opened 2026-04-19
Support loading Rapier native library via -Dsable_rapier_path=path

MovTery:
Support loading Rapier native library via `-Dsable_rapier_path=path`

For users on unsupported systems, this leverages Rust's cross-platform compilation capabilities, allowing them to build and load a custom native library for their specific environment.


CyberIvanbt:
Android?

PR #326, opened 2026-04-21
Fix Iris matrices when rendering sublevels

djefrey:
Iris replaces `ShaderInstance` with `ExtendedShader`.
When `ExtendedShader.apply()` is called, `modelViewInverse` and `normalMatrix` are updated using `MODEL_VIEW_MATRIX` value.

However, in `VanillaChunkedSubLevelRenderData.renderChunkedSubLevel()`, `MODEL_VIEW_MATRIX.set()` is called after `apply()` has already been called. This leaves the Iris matrices in an incorrect state.

This can be fixed by delaying the `apply()` call until just before rendering.

Before:

After:

### DIFF
--- a/buildSrc/src/main/groovy/multiloader-common.gradle
+++ b/buildSrc/src/main/groovy/multiloader-common.gradle
@@ -93,6 +93,7 @@ repositories {
 dependencies {
     compileOnly "maven.modrinth:lithium:mc1.21.1-0.15.3-neoforge"
     compileOnly "maven.modrinth:sodium:$sodium_version"
+    compileOnly "maven.modrinth:iris:$iris_version"
     compileOnly "maven.modrinth:distanthorizons:$distant_horizons_version"
     compileOnly("cc.tweaked:cc-tweaked-$minecraft_version-forge:$cc_tweaked_version")
     compileOnly("io.github.mortuusars.exposure:exposure-${minecraft_version}-neoforge:${exposure_version}") { transitive = false }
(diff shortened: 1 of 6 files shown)


ryanhcode:
tested, awesome! What's with the dynamic effects uniforms not needing to be uploaded anymore?

ryanhcode:
oh, is it because the apply is afterwards now and uploads them anyway? yeah okay

djefrey:
I made the change when I removed the now useless uploads. ~~However, as I'm reading the code again, I'm not sure if this is actually correct.~~

ryanhcode:
will this have any more noticeable impact than just uploading the individual uniforms like we were doing before?

djefrey:
I've done a simple benchmark: 354 sublevels on a flat world, no shaderpack.
This commit: ~760fps
With uploads readded: ~725fps
So it seems there's a small gain to prevent unnecessary uploads.

However, I noticed that this commit introduces a significant performance regression, as the previous version is around ~1400fps. This regression is caused by `apply()` reupload every uniforms, as the `dirty` flag in `Uniform` is ignored.
The only fix I see would be to compute and update manually `modelViewInverse` and `normalMatrix` if shader is an `ExtendedShader` instance.

ryanhcode:
With those numbers, I think it's worth it to have explicit compatibility with Iris here

djefrey:
This new commit replaces the delayed `apply()` approach with a direct compatibility with Iris.

With this new commit, there is no measurable performance impact with main.
When Complementary on High is used, previous commit ran at ~205 fps while this one achieves ~450 fps. There is no measurable performance impact with main in this situation either.

As for the extension itself, I'm unsure about the `sable$refreshModelViewDerivedMatrices` name, especially as a non native speaker. I'm open to suggestions if you have a better name.

ryanhcode:
thank you very much!

PR #341, opened 2026-04-21
Russian language

kucer0043:


VladisCrafter:
Firstly, #158 already exists. Secondly, ignoring Ё equals blatant inconsistency with translations of Minecraft, Create and basically everything else related to both.

PR #365, opened 2026-04-22
Fix for #48 (blocks duping upon assembly/disassembly)

### LINKED ISSUE #48: Item duplication bug from converting physics objects to blocks
### Description

Discovered running Create: Aeronautics

When a physics object with attached blocks such as redstone dust, levers, or buttons, is turned back into regular blocks, the attachments "break", dropping items and producing particles, but remain attached to the final structure.

Steps to recreate (using create: aero)
1. Place Swivel Bearing.
2. Place Block onto Bearing, and redstone dust or similar item onto Block.
3. Glue everything (other then the bearing) together and right click the bearing twice.

Sorry if this description is too vague or if this is a Aeronautics specific issue, I've never done this before and just want to help!

### Game Log

https://mclo.gs/KRG4edW

aphom:
### Rudimentary fix for brittle blocks duping upon assembly and disassembly.

Closes #48 

The old blocks that were being assembled / disassembled are first temporarily replaced by barriers to prevent brittle blocks from breaking prematurely, then they are replaced by air to complete the assembly / disassembly.

### Before the fix is applied
create_aeronautics_before_fix_low

### After the fix is applied
create_aeronautics_after_fix_low


aphom:
There are still issues with these blocks listed below that allow placement of brittle blocks, as they also have the "wrench_pickup" tag, so they mess up the sorting because they are included in the list of brittle blocks.. so the dupe bug still happens with them. Not sure how to fix this as of right now without manually making a specific list of blocks that should be excluded from the brittle blocks list, which would probably not be feasible because of other addons' blocks.

- Redstone Lamp, Target Block, Observer, Hopper, Pistons, Daylight Detector, Cart Assembler, Block of Industrial Iron, Block of Weathered Iron, Andesite Bars, Brass Bars, Copper Bars

From some testing it seems that these blocks mostly have issues while disassembling.

***This is no longer an issue with the new fix implementation.**

aphom:
With the new version of the fix, the issue with those select blocks that were mentioned previously no longer happens. Physics contraptions assemble and disassemble normally, without item duping.

create_aeronautics_after_fix_2

VoidLeech:
Wouldn't this still have issues with brittle blocks that are a little more complex than 'needs a sturdy face', such as cocoa blocks?

aphom:
> Wouldn't this still have issues with brittle blocks that are a little more complex than 'needs a sturdy face', such as cocoa blocks?

Those kind of blocks seem to work properly while this fix is applied:

https://github.com/user-attachments/assets/c5d64b7a-5654-447c-95f7-873c4dc54fea

dewayx:
I don't wish to be nagging, but may I ask purely out of curiosity, in what update this fix will be?

Ocelot5836:
The root cause of this issue is sub-levels of zero mass are destroyed and drop their items. Depending on the order the blocks are disassembled in this can lead to some parts of a sublevel being split off and then destroyed in a single tick. A better solution is being worked on that solves the root cause rather than a patch to prevent the symptoms.

Issue #247, opened 2026-04-20
Physics not affected by cobwebs


conmom456:
### Description

Physics objects will travel through cobwebs as if they don't exist at all.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/6519ea40-7a20-4cfe-9647-d39b411744ee" />

### Game Log

https://mclo.gs/Ng3dFV1

PR #393, opened 2026-04-22
Data-driven sublevel collisions with fluids

### LINKED ISSUE #246: Physics not affected by bubble columns
### Description

Pretty self explanatory. Dropping a physics object into a bubble column has no perceivable difference to normal water.

Image

### Game Log

https://mclo.gs/2wLdDTc

squigglesdev:
This PR gives fluids two properties: `buoyancy` and `viscosity`. This allows for any sublevels that come into contact with these fluids to behave more "realistically". Closes #246.

| Before | After |
| --- | --- |
|  |  |

It adds 4 more block tags, along with their corresponding physics block properties:
`#sable:low_buoyancy_fluid`
`#sable:high_buoyancy_fluid`
`#sable:runny_fluid`
`#sable:viscous_liquid`

(Bubble columns are hard-coded.)


Random-Scientist (changes requested):
Please remove the binary changes from this PR; we only commit natives built by GHA to avoid a contribution of malicious or non-reproduceable binaries. When your PR is merged the natives will be built by GHA and updated automatically.

squigglesdev:
I have also discovered with these new tags that while admittedly they are not technically a fluid, the `#sable:viscous_fluid` tag can be applied to cobwebs:



Leaving out a buoyancy related tag means that even this light block of oak planks won't float up to the surface like it would on water.
This would close #247, but might need to be tweaked to be perhaps part of an even more viscous group - cobwebs drag on vanilla entities significantly more than this.

Easyidle123:
Could this work per-dimension? I feel lava should be a viscous fluid in the overworld, but runny in the nether, as shown by its flow speed and distance in each dimension.

squigglesdev:
It absolutely could be, but I think the tags would start getting a bit verbose... `#sable:viscous_fluid_nether`, `#sable:runny_fluid_overworld`, etc.

larfy298:
I think that for the sake of having shorter tags and minimizing the amount that have to be used, `#sable:molten_fluid` and `#sable:sticky_fluid` would be the most efficient solution to the cobweb and lava problems, is it possible to have one tag (`molten_fluid`) function differently in two dimensions?
I also think this highly viscous property should be applied to honey blocks as they do not cover a full block with their collision and will slow things down on the border, even if the difference may only be minimal

squigglesdev:
That sounds like a great way of handling it. I'll take a look into implementing that. 

Also I totally agree with the honey block idea!

squigglesdev:
I've added a `#sable:molten_fluid` tag that allows fluids to be viscous in the overworld and end, whilst runny in the nether. Unfortunately giving honey blocks the `#sable:viscous_fluid` tag completely removes their collision so I wasn't able to do so.

TechTastic:
This is neat!

PR #404, opened 2026-04-22
Add some operator utility items for sublevels

whimsicalwren:
adds four items to the operator utilities tab:
sublevel assembler: assembles the clicked block into a sublevel
sublevel area assembler: assembles all blocks between two clicked positions into a single sublevel
single block sublevel area assembler: assembles all blocks between two clicked position into any number of single-block sublevels
sublevel remover: removes the clicked sublevel
i may work on adding disassembly items later, however that's slightly more complicated as far as i can tell and not absolutely neccessary.

this pr works both indev and out-of-dev, on neoforge and fabric.

Screenshot 2026-04-22 131550
Screenshot 2026-04-22 131554
Screenshot 2026-04-22 131557
Screenshot 2026-04-22 131600


ryanhcode:
Unfortunately I think adding actual items is past the scope of Sable, I'd rather it stay purely a library mod

You may be adjacently interested in Touys by Kyan, also: https://modrinth.com/mod/touys
It adds a Bubble Blower item that allows for easy area selection, and an Evil Orb that deletes sub-levels alongside some other operator utilities.

whimsicalwren:
that's fair, i might look into that mod and submit a pr to that and add some of these items

Issue #394, opened 2026-04-22
UDP Server Spamming Logs


ashtr4l:
### Description

Poyoraz and I have been investigating an issue present when running the mod on Linux servers causing packets to flood the console and cause lag spikes.
 
These packets are queries for server metadata (playercount, motd, etc) used by the server list (or server hosts) to display current information about the server.
 
The reason that these packets are throwing errors is due to Minecraft's Query system from QueryThreadGs4 not initialising correctly, the Epoll system used on Linux cannot handle concurrent access to the same port by multiple services without additional efforts (Namely the SO_REUSEPORT option on channels). The QueryThreadGs4 Class does not have this measure.
 
As the UDP server is started first and acquires access to the port, the Query system is prevented from acquiring access. All packets that should be caught and handled by this service are instead picked up by the UDP server, which throws errors as it does not recognise what to do with them.

(The query system not initialising can be seen at line 481 in the attached logs)
 
The two potential fixes would be:
- Force Minecraft's query system to use the SO_REUSEPORT option.
    - Overriding base Minecraft methods would not be ideal for compatability.
- Move the UDP server initialisation so that it occurs after the query thread and enable the SO_REUSEPORT option on the channels it opens.
    - In the current UDP Server code, enabling the flag would be achieved by inserting ".option(EpollChannelOption.SO_REUSEPORT, true)" into ServerConnectionListenerMixin.java at lines 70 & 96.

### Game Log

https://mclo.gs/fbdtbji

Poyoraz:
For any end users who are coming across this problem, [Connectivity](https://www.curseforge.com/minecraft/mc-mods/connectivity) does fix this issue.

Poyoraz:
https://github.com/netty/netty/issues/5799 SO_REUSEPORT also seems to not be available in Bootstrap, although in our tests we didn't get the mentioned error message so some stuff might've changed in the last 10 years

Ocelot5836:
Does `fix/udp` resolve the issue?

Poyoraz:
Nope, we already tried the same solution in that branch sadly

ashtr4l:
We tested SO_REUSEPORT while exploring the solutions to the bug, that's where I got the knowledge of the line positions to insert the code. 

SO_REUSEPORT only works to allow access to a port that is already claimed, the UDP server acquires its port access first, so just adding that option doesn't change anything.

The whole code for initialising the UDP server would need to also be moved / delayed so that it executes after the query system is initialised.

[...]

PR #406, opened 2026-04-22
add the ability to control udp port from config

Poyoraz:
to fix the following issue:

https://github.com/ryanhcode/sable/issues/394

The server should probably communicate what the udp port is instead of assuming the client knows, but I don't know enough about the internals of the project to do that


Trzezwyy:
> to fix the following issue:
> 
> #394
> 
> The server should probably communicate what the udp port is instead of assuming the client knows, but I don't know enough about the internals of the project to do that

Hi, could you share a compiled build of this PR? I'm running a dedicated server and need this fix

Poyoraz:
> > to fix the following issue:
> > #394
> > The server should probably communicate what the udp port is instead of assuming the client knows, but I don't know enough about the internals of the project to do that
> 
> Hi, could you share a compiled build of this PR? I'm running a dedicated server and need this fix

Sorry, I don't redistribute the mod, you'll have to build the rust natives and compile the mod yourself

nick9509:
when that PR will be merged?

PR #426, opened 2026-04-23
Update sable companion to 1.6.0

Ocelot5836:
Adds support for rectilinear distance (the maximum of x, y, or z distance). Some mods prefer this distance over cartesian distance.

**This requires sable companion `1.6.0` to be published to the maven before it can be merged**


ryanhcode:
awesome1

PR #449, opened 2026-04-23
WIP physics tracing/recording

Random-Scientist:
implements a tracing system that serializes all sable JNI commands to a file for playback by a minimal runner binary.
this opens up a lot of exciting possibilities for improving the physics engine down the line.
* integration testing and validation of changes
* ease of profiling/measuring performance changes on real recorded workloads
* repeatable benchmarks
* capturing PGO instrumentation data in CI builds
* faster iteration times to test physics changes
* reliable and controlled reproducers for physics bugs

At present there is a deadlock issue when recording with reentrant native calls. Also, this doesn't support (tracking the behavior of) external material collision handler callbacks yet.


Random-Scientist:
Intend to rebase and merge this after #161 because unifying all of the vector types will make state serialization significantly less painful

PR #471, opened 2026-04-24
Add needed Repositories to Home.md

EyaeM:
Me and a bunch of friends were having issue's with importing Sable. And found that these repositories were needed for gradle to accept Sable.

### DIFF
--- a/wiki/Home.md
+++ b/wiki/Home.md
@@ -64,6 +64,15 @@ repositories {
             includeGroup("dev.ryanhcode.sable-companion")
         }
     }
+
+    maven { //Veil
+        url "https://maven.blamejared.com"
+    }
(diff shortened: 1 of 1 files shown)


MarbleGateKeeper:
Unnecessary. You should already have these repositories on your buildscript when developing Create addon.

EyaeM:
> Unnecessary. You should already have these repositories on your buildscript when developing Create addon.

Sable is not only for create add-ons. What if someone is developing something unrelated to create using sable?

MarbleGateKeeper:
> Sable is not only for create add-ons. What if someone is developing something unrelated to create using sable?

Make sense. I'll hide my reply as off-topic.

Issue #431, opened 2026-04-23
Sable hard crashing the JVM


devprinter:
### Context

The log file ended abruptly and there was no crash report. The only evidence of this happening was in stdout of the server

Apparently this has been happening semi regularly, about once per day

I am using a slightly outdated version of sable, but I'm posting it anyway in case you guys didn't know about it and it's still present using the current version. Haven't had a chance to update yet

create versions:
sable-neoforge-1.21.1-1.0.6.jar
create-aeronautics-bundled-1.21.1-1.0.3.jar
create-1.21.1-6.0.9.jar

All mods:

<img width="366" height="716" alt="Image" src="https://github.com/user-attachments/assets/7cb53a39-7e6e-491f-a0d5-e43f7b574cea" />

Server specs:
<img width="586" height="352" alt="Image" src="https://github.com/user-attachments/assets/d6c69ee6-5e6e-4a55-959f-19228bfa277b" />

### Crash Report

https://gist.github.com/NeoCamelot/8ec38b96d94f59d5bc6b942715a53add

Random-Scientist:
I believe is a known issue regarding sublevels breaking fragile blocks in the world
`Internal error: island should be awake.`

devprinter:
It just happened again with these versions:
sable-neoforge-1.21.1-1.1.3.jar
create-1.21.1-6.0.10.jar
create-aeronautics-bundled-1.21.1-1.1.3.jar

PR #477, opened 2026-04-24
Catch Rapier panics at FFI boundary to prevent crash

hexadexxa:
Rapier internal assertions can panic during the physics step. Since the
JNI functions use `extern "system"`, this unwinds across the FFI boundary
and aborts the JVM.

Wraps `pipeline.step()` and `tick()` in `catch_unwind` so panics are
caught and the tick is skipped instead of crashing.

### DIFF
--- a/common/src/main/rust/rapier/src/lib.rs
+++ b/common/src/main/rust/rapier/src/lib.rs
@@ -428,18 +428,20 @@ pub extern "system" fn Java_dev_ryanhcode_sable_physics_impl_rapier_Rapier3D_tic
     scene_id: jint,
     _time_step: jdouble,
 ) {
-    unsafe {
-        if let Some(state) = &mut PHYSICS_STATE {
-            rope::tick(scene_id);
-            joints::tick(scene_id);
+    let _ = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+        unsafe {
+            if let Some(state) = &mut PHYSICS_STATE {
+                rope::tick(scene_id);
+                joints::tick(scene_id);
 
-            let Some(scene) = state.scenes.get_mut(&scene_id) else {
-                panic!("No scene with given ID!");
-            };
+                let Some(scene) = state.scenes.get_mut(&scene_id) else {
+                    return;
+                };
 
-            compute_buoyancy(scene);
+                compute_buoyancy(scene);
+            }
         }
-    }
+    }));
 }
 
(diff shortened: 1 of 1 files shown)


hexadexxa:
Should also fix issue #431 if it's the same issue

Random-Scientist:
this is a really bad idea. the rust panics are panics for a reason; they indicate a fatal error in the physics engine (there is no mechanism for handling recoverable errors at present), ignoring these fatal errors will just invite more UB or breakages in later calls or physics steps, making everything harder to debug.

ryanhcode:
i think it still propagates to a server crash, just through a jvm exception

Random-Scientist:
> i think it still propagates to a server crash, just through a jvm exception

At present it simply seems to ignore panics altogether (hence my comments). Clearing the physics state and raising a JVM exception with a rust backtrace would be nice but the present version of the pr just makes debugging native crashes harder.

ryanhcode:
yeah, ignoring the panics is an absolute no-go. I'd be happy if we raised a JVM exception with the details from the panic and unrecoverably crashed the server

hexadexxa:
I see, I'll try doing that later

hexadexxa:
Updated. Now catches the panic and throws a RuntimeException with the panic message via env.throw_new, so the server crashes with a proper crash report

Random-Scientist (approved):

common/src/main/rust/rapier/src/lib.rs
```diff
@@ -421,31 +421,53 @@ pub extern "system" fn Java_dev_ryanhcode_sable_physics_impl_rapier_Rapier3D_ini
...
-            rope::tick(scene_id);
-            joints::tick(scene_id);
+    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+        unsafe {
+            if let Some(state) = &mut PHYSICS_STATE {
+                rope::tick(scene_id);
+                joints::tick(scene_id);
 
-            let Some(scene) = state.scenes.get_mut(&scene_id) else {
-                panic!("No scene with given ID!");
```
Ocelot5836, line 437: This is supposed to panic because something on the java side went horribly wrong. If someone passed in the wrong scene id it needs to fail

Ocelot5836 (changes requested):

Ocelot5836 (approved):
LGTM

PR #497, opened 2026-04-25
Modded Weight Compat

Ad3lineL:
Fixed two missing tags and tagged a few modded blocks from:
Decorative Blocks
Woodworks
Quark
Supplementaries
Architect's palette
Storage Drawers
Another Furniture
Farmer's Delight
No Mans Land


tazercopter:
good pr, would merge.

tazercopter:
I would re-add the #c:ladders as an optional tag, it's likely that mods such as woodworks use it which is why addy added them.

ryanhcode:
hi addy

Ad3lineL:
hi ryan

PR #520, opened 2026-04-25
Little Tiles Partial Compatability

milkev:
**The Issue I'm Trying to Fix**
Little Tiles incompatability. It [crashes](https://pastebin.com/ujvkkUwt) when physics objects have little tiles on them and an entity collides with the physics object. Notably no crash occurs when physics objects collide with little tiles in the "real world", they just simply treat the little tiles as if they dont exist.

**What I Did**
When [VoxelShapeMixin.sable$allBoxes](https://github.com/ryanhcode/sable/blob/main/common/src/main/java/dev/ryanhcode/sable/mixin/voxel_shape_iteration/VoxelShapeMixin.java#L23) is called in a context relevant to little tiles, it will always be null. I added some null checks for this in entity collision, and just skip over the relevant code if it is null. 

**The Results**
This results in little tile blocks being placeable and visible on physics objects, however little tile blocks are completely ignored for physics object collision checks. (same as when a physics object collides with little tile blocks in the "real world")

I have not tested little tiles advanced functionality as I dont have experience working with that stuff in the first place, I just make visual stuff.

Little Tiles on physics objects do not render upon first entering a world, requiring a little tile to be modified on the physics object to make little tiles on that physics object visible again.

Little Tiles are also not highlightable or breakable by hand while on physics objects. They can still be modified by little tiles tools and premade structures can be placed.

Little Tiles in the "real world" are not affected at all (from my testing) by these changes.

**Other Options**
Option 1: A better (but still easier than full compatability) option would be to prevent Little Tiles from being placed on physics objects in the first place. However, doing so via preventing placement via [canPlace](https://github.com/ryanhcode/sable/blob/main/common/src/main/java/dev/ryanhcode/sable/mixin/block_placement/BlockPlaceContextMixin.java#L79) doesnt work due to LittleTiles placement logic not obeying canPlace. (And I'm not smart enough to figure out how their placement logic works). If there is a way to prevent certain blocks or block types from existing on Sub-Levels on a deeper level that would be better (and honestly a great addition, gonna make a relevant issue for this) 

Option 2: Instead of skipping the relevant code, generate a full-box collision for little tiles. Depending on opinion, this may be better than them having no collision. I'm whatever either way I just dont know how to do this.

### DIFF
--- a/common/src/main/java/dev/ryanhcode/sable/mixin/voxel_shape_iteration/VoxelShapeMixin.java
+++ b/common/src/main/java/dev/ryanhcode/sable/mixin/voxel_shape_iteration/VoxelShapeMixin.java
@@ -31,6 +31,12 @@ public abstract class VoxelShapeMixin implements FastVoxelShapeIterable {
     @Override
     public Iterator<BoundingBox3dc> sable$allBoxes() {
         synchronized (this) {
+            
+            if(sable$boxIterator == null) {
+                return null;
+                //This is null when this mixin is relevant to little tiles
+            }
+            
             final long id = Thread.currentThread().threadId();
             FastVoxelShapeIterator iterator = this.sable$boxIterator.get(id);
 
--- a/common/src/main/java/dev/ryanhcode/sable/sublevel/entity_collision/SubLevelEntityCollision.java
+++ b/common/src/main/java/dev/ryanhcode/sable/sublevel/entity_collision/SubLevelEntityCollision.java
@@ -275,26 +275,29 @@ public class SubLevelEntityCollision {
                         }
 
                         final Iterator<BoundingBox3dc> iterator = ((FastVoxelShapeIterable) voxelShape).sable$allBoxes();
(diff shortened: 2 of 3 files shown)


hexadexxa:
Surely this would be something good to have for all sorts of weird mod compatabilities

Ocelot5836:
This happens because creative core violates the java contract by directly allocating voxel shapes for some reason. It also goes against the idea of voxel shapes being immutable. [This](https://github.com/CreativeMD/CreativeCore/blob/dc3351d38c37098b6fd5d1290089328a9bcc9bde/src/main/java/team/creative/creativecore/common/util/math/box/BoxesVoxelShape.java#L31) should be fixed on their side

CreativeMD:
@Ocelot5836 I get your point. The reason why I do these weird stuff because performance was extremely bad. VoxelShapes are horrible when there are 20+ boxes in one block.

Currently I'm busy with other stuff, but I plan to take a good look at compatibility and try to fix stuff from my side.

PR #546, opened 2026-04-26
Replace most overwrite/redirect with wrapMethod(s)

cylian914:


ryanhcode:
Apologies, I'm not interested in approving changes this sweeping, and I agree with a number of these cases being redirects/overwrites instead of wrapoperations / wrapmethods.

cylian914:
would it be good if i separate the overwrite from the redirect and put the redirect into multiple pr?

ryanhcode:
I would rather target specific cases where the ways that individual mixins are being done are problematic

cylian914:
most usage of non chainable injection is problematic since its made to be a lib, to work with other mods
overwrite that copy paste part of vanilla code are fine as other mod injecting will do it in the overwrite (and if they overwrite too it will replace it), so i can omit it if you want, but 2 mod can't redirect the same injection point as it will often crash

all redirects that can be changed for wrapMethods are problematic
if you don't want to have a giant commit with all changes i can always separate it but the issue is still there and it still need to change

PR #548, opened 2026-04-26
Add support for uploading artifacts.

pizzapi2012:
Would be nice for git commits before they release.


ryanhcode:
I'm not interested in making a change like this, commits and the main branch are used for development and I don't want to publish jars for them regularly at the moment.

pizzapi2012:
> I'm not interested in making a change like this, commits and the main branch are used for development and I don't want to publish jars for them regularly at the moment.

ok

PR #549, opened 2026-04-26
Fix #128

### LINKED ISSUE #128: Dynamic Light issue
### Description

Sodium Dynamic Lights/LambDynamicLights do not illuminate entities if Sable is installed
Version: NeoForge 1.21.1
Sable: 1.0.5
Image

### Game Log

https://mclo.gs/psAAh6S

Kilip1000:
Fixes #128.
This is done by taking the maximum of the light level applied in the method, and the original light level with no changes applied.

### DIFF
--- a/common/src/main/java/dev/ryanhcode/sable/mixin/entity/entity_rendering/EntityRendererMixin.java
+++ b/common/src/main/java/dev/ryanhcode/sable/mixin/entity/entity_rendering/EntityRendererMixin.java
@@ -1,5 +1,6 @@
 package dev.ryanhcode.sable.mixin.entity.entity_rendering;
 
+import com.llamalad7.mixinextras.injector.ModifyReturnValue;
 import dev.ryanhcode.sable.Sable;
 import dev.ryanhcode.sable.api.entity.EntitySubLevelUtil;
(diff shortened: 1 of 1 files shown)


ryanhcode:
we would then be doing double the light queries

Kilip1000:
> we would then be doing double the light queries

Do you mean to say that preserving the original light-level is not worth it, unless the execution of those light queries is at no additional cost? I, frankly, believe that's infeasible/impossible to do.

ryanhcode:
No, there is a way to both preserve the original light level here and also to incorporate the sub-level light levels

Kilip1000:
> No, there is a way to both preserve the original light level here and also to incorporate the sub-level light levels

How's that possible if the original value does not get evaluated/incorporated?

ryanhcode:
also, calling Math.max on an already packed light value pair isn't a valid operation. I'm working on a fix that shouldn't add any more additional queries and should fix this

Kilip1000:
Ah, right, I forgot one of the light values is a higher power of two than the other and would thus unintentionally take priority!
I'm sorry for wasting your time 😓

ryanhcode:
you're all good! thank you for bringing the issue to light

PR #568, opened 2026-04-27
Uses wrapOperation instead of calling the redirected function

cylian914:
this doesn't contain any changed logic, just replace redirected function call, so it shouldn't take too much time to validate

The usage of redirect is problematic since it doesn't chain with other mod, keeping them will cause issue


ryanhcode:
I'm not really interested in approving changes these sweeping, and I disagree with some of these being wrapoperations- are you having issues or incompatibilities due to any of these mixins in particular?

cylian914:
there isn't any issue yet but they will appear if that much redirect stays

Also with what wrap operation do you disagree exactly? Because all the one this commit change are the one who were calling the redirected method inside of it (often with the exact same parameter)

PR #585, opened 2026-04-28
Fix #571

### LINKED ISSUE #571: UnsupportedOperationException: Cannot change blocks in nonexistent plot holder — race condition between Sable contraption disassembly and Create Deployer
### Context

Environment:
Minecraft: 1.21.1
NeoForge: 21.1.228
Sable: 1.1.3
Create: 6.0.10
Server type: Dedicated, ~30 concurrent players

Description
The server crashes repeatedly with UnsupportedOperationException: Cannot change blocks in nonexistent plot holder when two specific operations happen simultaneously:

- A Sable physical contraption is disassembling back into blocks using an simulated:assembler
- A Create Deployer places a block in a nearby chunk at the same moment

This appears to be a race condition in Sable's plot holder management - the plot holder exists for one operation but is in a transitional state by the time the second operation attempts to notify it, causing the crash. The issue has been confirmed to persist without c2me installed.
The crash happens consistently on a server with 50 players because both Sable contraptions and Create Deployers are in active use at all times.
The crash is not triggered by either operation in isolation. It requires both happening at the same time.
Steps to Reproduce

- Have an active Create Deployer running continuously (e.g. an XP farm)
- Have a player disassemble a Sable physical contraption back into blocks nearby
- Server crashes with the exception below

The crash report does not provide relevant information as the server closes because of a tick being too long, and the issue happens before the crash but leads to it.

### Crash Report

Key crash line java.lang.UnsupportedOperationException: Cannot change blocks in nonexistent plot holder   at ServerChunkCache.handler$sable$blockChanged.

justlofe:
Fixes [this](https://github.com/ryanhcode/sable/issues/571) one issue. Cancels the update instead of causing an exception.

### DIFF
--- a/common/src/main/java/dev/ryanhcode/sable/physics/impl/rapier/Rapier3D.java
+++ b/common/src/main/java/dev/ryanhcode/sable/physics/impl/rapier/Rapier3D.java
@@ -648,6 +648,7 @@ public class Rapier3D {
     @ApiStatus.Internal
     public static native void configMinIslandSize(int islandSize);
 
+    @ApiStatus.Internal
     public static native void dispose();
 
     @ApiStatus.Internal


ryanhcode:
This feels like more of a bandaid- blocks should not be modified in deleted plots, and if they are, we should fix that root issue

justlofe:
I'll try to fix the main problem and ping you back when it's done.

hotpoc:
> This feels like more of a bandaid- blocks should not be modified in deleted plots, and if they are, we should fix that root issue

@ryanhcode I know this is closed, but why wouldn't this PR work? Right now, `blockChanged()` in `ServerChunkCacheMixin` treats getting `null` from `container.getChunkHolder(pos)` as an illegal state, while `SubLevelContainer#getChunkHolder()` is `Nullable`. If not cancelling the call, shouldn't it just default to vanilla behavior?

PR #632, opened 2026-04-29
Improve assembly block handling to prevent item duplication

SAILQAQ:
Two fixes combined:
1. From https://github.com/ryanhcode/sable/pull/365: Replace old blocks with barriers before removal. This prevents brittle blocks (torches, buttons, etc.) from breaking when their supportblock is removed, which could also cause duplication.

2. New fix: After saving BE NBT, call level.removeBlockEntity() to detach the block entity from the world before it can drop items. Then destroy barriers via level.setBlock() with UPDATE_MOVE_BY_PISTON | UPDATE_SUPPRESS_DROPS | UPDATE_CLIENTS — the same approach Create uses in its contraption system.

### DIFF
--- a/common/src/main/java/dev/ryanhcode/sable/api/SubLevelAssemblyHelper.java
+++ b/common/src/main/java/dev/ryanhcode/sable/api/SubLevelAssemblyHelper.java
@@ -27,8 +27,6 @@ import net.minecraft.core.SectionPos;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.server.level.FullChunkStatus;
 import net.minecraft.server.level.ServerLevel;
-import net.minecraft.world.Clearable;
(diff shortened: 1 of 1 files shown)


SAILQAQ:
This should fix the duplication issues for most modded containers (those that don't implement Clearable) and attachable blocks.

VoidLeech:
Just coming over here because it got mentioned on the Create github, but why is this taking another PR when the secondary fix seems to be completely standalone? I've some small concerns but I'll drop those on the original.

SAILQAQ:
> Just coming over here because it got mentioned on the Create github, but why is this taking another PR when the secondary fix seems to be completely standalone? I've some small concerns but I'll drop those on the original.
> 

That makes sense, I'll close this PR and open a standalone one with just the `removeBlockEntity` fix.

Issue #510, opened 2026-04-25
Sable implementation of `getEntities` causing decreased performance


MrSeven3:
### Description

Sable appears to change part of a function that gets the entities in a world (makes sense), but for some reason, even when there are no sublevels on the server (to my knowledge), the part that looks for sublevels is taking obscene amount of time per tick. This server has a high number of chunks and entities, but for some reason, part of the collision checks for items are take over 50ms per tick. Buried in the collision check, it calls `getEntities`, for which the vanilla function returns in less than 0.1ms, but the Sable function takes over 50 ms

<img width="621" height="104" alt="Image" src="https://github.com/user-attachments/assets/88190ea0-d0bd-4107-ba67-09705a904f66" />

[spark profiler](https://spark.lucko.me/7hBhHSXADi)

### Game Log

https://mclo.gs/ilM23sC

PR #633, opened 2026-04-29
perf: add per-tick occupied-chunk cache to short-circuit entity queries

Rail1bc:
## Summary

This PR adds a **per-tick occupied-chunk cache** that enables O(1) lookups for "does this chunk contain a sub-level?" queries. The cache is consumed by `SubLevelInclusiveLevelEntityGetter` to short-circuit the expensive sub-level traversal when the queried AABB clearly does not intersect any sub-level at all.

## Background

Issue #510 reports that `SubLevelInclusiveLevelEntityGetter.get()` is extremely slow (\>50ms per call vs \<0.1ms for vanilla) because every invocation of `get(AABB)` iterates over **all** loaded sub-levels — even when those sub-levels are nowhere near the query region. On servers with many chunks and entities this adds up to severe tick-time regression.

### `SubLevelContainer.java`
- Added a **`LongSet occupiedChunks`** field — a set of `ChunkPos.asLong()`-encoded chunk positions that are covered by at least one loaded sub-level.
- The cache is **rebuilt from scratch every tick** in `tick()` so that moving/rotating sub-levels are tracked correctly.
- Added **`isChunkOccupied(chunkX, chunkZ)`** — public O(1) query method backed by the cache.

### `SubLevelInclusiveLevelEntityGetter.java`
- Added a **`hasSubLevelInAABB(AABB)`** helper that performs a two-level short-circuit:
  1. **Dimension-level:** if the world has zero loaded sub-levels, skip entirely.
  2. **Chunk-level:** convert the AABB to a chunk range and probe the occupied-chunk cache.
- Both `get(AABB, Consumer)` and `get(EntityTypeTest, AABB, ...)` now **return early via the vanilla delegate** when no sub-level intersects the query region.

## Performance Impact

| Scenario | Before | After |
|----------|--------|-------|
| Empty region (no sub-levels nearby) | \>50ms (full traversal) | \<0.1ms (vanilla) |
| Region with sub-levels | \>50ms | \>50ms + \<0.001ms cache probe |
| Dimensions without any sub-levels | \>50ms | \<0.1ms |

The cache rebuild itself costs O(m·s) per tick (m = sub-level count, s = chunks per sub-level), which is negligible compared to the O(n·m) it eliminates across the thousands of `get(AABB)` calls per tick.

### DIFF
--- a/common/src/main/java/dev/ryanhcode/sable/api/sublevel/SubLevelContainer.java
+++ b/common/src/main/java/dev/ryanhcode/sable/api/sublevel/SubLevelContainer.java
@@ -11,6 +11,8 @@ import dev.ryanhcode.sable.sublevel.plot.PlotChunkHolder;
 import dev.ryanhcode.sable.sublevel.storage.SubLevelOccupancySavedData;
 import dev.ryanhcode.sable.sublevel.storage.SubLevelRemovalReason;
 import dev.ryanhcode.sable.util.iterator.ListBackedFilterIterator;
+import it.unimi.dsi.fastutil.longs.LongOpenHashSet;
+import it.unimi.dsi.fastutil.longs.LongSet;
 import it.unimi.dsi.fastutil.objects.ObjectArrayList;
 import it.unimi.dsi.fastutil.objects.ObjectList;
 import net.minecraft.client.multiplayer.ClientLevel;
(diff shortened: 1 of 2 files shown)


Rail1bc:
Hi @ryanhcode, thanks for taking a look at this PR.

I noticed it was closed without any comment or review. Could you share what concerns you have about this approach? I'd be happy to adjust the implementation if there are specific issues with the design or code quality.

I understand you may be busy — just wanted to understand if there's something fundamentally wrong with this direction so I can learn and improve. Thanks!

PR #650, opened 2026-04-30
feat: add loongarch64 (LoongArch) architecture & riscv64 (RISC-V) arc…

OvOla2:
…hitecture support

This is a fresh start. The following changes enable cross-compilation and runtime support for LoongArch64 and RISC-V 64-bit architectures.

Changes:

1. build.gradle
   - Add loongarch64 and riscv64 to supportedTargets
   - Override glibc version: use 2.36 for loongarch64 & riscv64, keep 2.17 for others

2. Rapier3D.java
   - Add architecture detection for loongarch64 (os.arch starts with loongarch64 or loong64)
   - Add architecture detection for riscv64 (os.arch starts with riscv64 or riscv64gc)
   - Map them to native library names: sable_rapier_loongarch64_linux.so, sable_rapier_riscv64_linux.so

3. Dockerfile (container/zigbuild)
   - Add rustup target: loongarch64-unknown-linux-gnu
   - Add rustup target: riscv64gc-unknown-linux-gnu

These modifications allow the project to build and run on Loongson (LoongArch64) and RISC-V 64-bit platforms. glibc 2.36 is explicitly set for these two architectures to satisfy toolchain requirements.

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -19,7 +19,8 @@ def mac(arch) {
 }
 
 def linux(arch) {
-    [triple: "${arch}-unknown-linux-gnu", suffix: ".2.17", arch: arch, os: "linux", ext: "so", lib: true]
+    def glibcVersion = (arch in ["loongarch64", "riscv64gc"]) ? ".2.36" : ".2.17"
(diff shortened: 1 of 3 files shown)


IThundxr:
There aren't really enough people using loongarch64 and riscv to value shipping these natives by default, i've got some ideas for how to handle some more obscure platforms and architectures that i'll explore later.

PR #659, opened 2026-04-30
Fix item duplication using same block removal strategy as Create contraptions

### LINKED ISSUE #683: Minecolonies Rack Duplication
### Description

When assembling a physics objects, all minecolonies racks drop their itens on ground, however, they still exists inside the rack

### Game Log

https://mclo.gs/xBmmn1r

SAILQAQ:
After saving BE NBT, call `level.removeBlockEntity()` to detach the block entity from the world before it can drop items. Then destroy old blocks via `level.setBlock()` with `UPDATE_MOVE_BY_PISTON | UPDATE_SUPPRESS_DROPS | UPDATE_CLIENTS` — the same approach Create uses in its contraption system.

This replaces the Clearable/RandomizableContainer approach which only covered mods that explicitly implement those interfaces.


Ocelot5836:
The correct way to fix this kind of duplication is to implement the `Clearable` interface from Minecraft. Many mods do not implement this interface correctly, so I proposed an alternate solution where blocks can be added to a "silent assembly" tag which has the same behavior as this PR. https://github.com/ryanhcode/sable/pull/981

PR #664, opened 2026-04-30
Add Mechanical Hand support for different sublevels

YassiGame:
## Summary

This pull request adds support for Mechanical Hands, allowing them to interact with both Sable sublevels and the normal Minecraft world at the same time. 

The goal is to make Mechanical Hands work correctly across both types of interactions without breaking input/output behavior or animations. 

I tested the changes, and everything seems to work pretty well so far.

Two demonstration videos are included below:

- Video 1: https://youtu.be/1WIbPLmrslc
- Video 2: https://youtu.be/Zm5AmhnCL0M

## Changes

- Added `MechanicalArmSublevelInteraction` mixin
  - Provides compatibility and range support for Mechanical Hands with sublevels.

- Added `ArmInteractionSubLevelPointAnimation` mixin
  - Fixes the Mechanical Hand animation when interacting with sublevel points.

- Added `MechanicalArmSubLevelInteractionPointHandler` mixin
  - Prevents input and output points from being removed when placing a Mechanical Hand, as long as the radius is valid with the sublevels.

- Updated the error message in the existing `MechanicalArmSublevelFailure` mixin
  - Improves the error feedback related to Mechanical Hand sublevel interactions.


Cyvack:
Mechanical arms not working across sublevels is intended explicitly. However, If we decide to visit the implementation in the future it will be done internally.

PR #696, opened 2026-05-01
Android binary support

Gogo64pro:
Add support for android rust binaries. *Probably* worth adding a new CICD pipeline so you dont build useless targets

*Screenshot of sable running on an android device with amethyst launcher*
Screenshot

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -30,6 +30,10 @@ def windows(arch) {
     [triple: "${arch}-pc-windows-msvc", suffix: "", arch: arch, os: "windows", ext: "dll", lib: false]
 }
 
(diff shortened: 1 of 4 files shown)


Gogo64pro:
Mb I accidentally clicked this copilot thing don't mind it

IThundxr:
no. we've already made it clear that sable won't ship android natives.

whitebelyash:
Android detection could be still useful though.

Kurizek27:
> no. we've already made it clear that sable won't ship android natives.

why? people already made it possible
https://github.com/sa1672ndo/sable-android-patch

PR #700, opened 2026-05-02
Fix crash caused by teleporting using Iron's Teleport scroll

### LINKED ISSUE #496: Teleportation onto physics objects is broken
### Description

Using iron's spells and spellbooks, or maybe other mods (not tested with other mods that add teleportation mechanics), causes unintended behavior. Sometimes it crashes the game, sometimes the inventory gets wiped after the restart, sometimes it just teleports to 0 0 0, in any case this is broken. 

Steps to reproduce: 
1. Have any physics object - a ship, an airplane, a single log with a physics assembler
2. Teleport onto it using teleportation spell 
3. Have fun!

PS - ender pearls work fine but modded teleports, at least from iron's spells, do not. 

### Game Log

https://mclo.gs/7Ut9UBk

gjorgdy:
The mixin for player teleportation did not cover all possible teleport methods the class contains. One of these methods which was not covered is called by Iron's Spells and Spellbooks to teleport using the teleport and the portal spells. 

A small other issue still persists, portal frames only being to teleport players **to** sublevels, but not detecting player collision when the frame is in a sublevel. I am not quite sure if this is an issue caused by Iron's or by Sable at this moment.

Closes #496

### DIFF
--- a/common/src/main/java/dev/ryanhcode/sable/mixin/entity/teleport_players/ServerPlayerMixin.java
+++ b/common/src/main/java/dev/ryanhcode/sable/mixin/entity/teleport_players/ServerPlayerMixin.java
@@ -14,6 +14,7 @@ import net.minecraft.world.level.Level;
 import org.joml.Vector3d;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.Unique;
 
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


Variapolis:
Is there a reason why your PR is for main rather than your feature/irons-compat branch? In my brief testing I was able to teleport without issue using a build based off of that branch, including with portal frames. It seems that it is a better fix for this issue.

I'm planning on testing it on a server with around 8 players over the next few days and would be happy to report any feedback if it is wanted.

gjorgdy:
I initially also made another PR with that "irons-compat" branch, but the Sable devs said they would prefer to not have to maintain support for Irons and that it should be fixed on that mod itself.

This PR is a generic fix to teleportation and should also fix other mods without requiring to add implicit compatibility.

Variapolis:
I'm seeing another issue here where the teleport using portal frames fails if the sublevel is too far way in an unloaded chunk. Is this possible to fix?

gjorgdy:
I have looked into it, but Sable does not seem to have a simple tool to calculate the relative position of an unloaded sublevel.

There have been other people who have gotten it to work tho, like [WaystonesSable](https://github.com/SShakusora/WaystonesSable).

Preferably this should be added to the utility functions in Sable Companion.

PR #708, opened 2026-05-02
Added more compatibility for Iron's Spells and Spellbooks

gjorgdy:
Besides the game-breaking crash, there were a few smaller issues when using Iron's spells and/or blocks in sublevels.

In this PR I fixed the following issues; 
- Portal frames detecting players entering when it's placed in a sublevel.
- Particles of the Touch Dig beam don't calculate all the way to the sublevel plot.

Issues which still exist:
- When teleporting from the world to a sublevel using a portal frame, the rotation of the sublevel is not accounted for.

Expands on #700


tazercopter:
it would be much better if iron's spells and spellbooks just implemented sable companion, as it was made for this exact reason and all your mixins are just adapting the mod to sable companion. having extra mixins for the mod in base sable means it has to be maintained whenever anything in the original mod changes.

gjorgdy:
No problem, I will close this PR and submit one for Iron's instead.

PR #726, opened 2026-05-03
Dimensional Teleportation

jbookout98:
This should add parents to sublevels. I also increased the sable teleport command to allow for dimensional teleportation. This should work for independent sublevels as well as family sublevels. 

There are issues with ropes teleporting in that the connected sublevels will spring around on teleporting and move random places. I've tried to counteract that by disabling velocity carry over but that did not fix it and I've just given up on ropes.

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -6,6 +6,7 @@ import java.nio.file.Files
 buildscript {
     dependencies {
         classpath 'at.yawk.lz4:lz4-java:1.11.0'
+
     }
 }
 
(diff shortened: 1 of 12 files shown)


enjoyerofthemusics:
Does this PR use AI? Some of the comments certainly look just like it lol

jbookout98:
> Does this PR use AI? Some of the comments certainly look just like it lol

Was definitely messing around with it.

enjoyerofthemusics:
Contributing says "The submitted code of your PR must be your own, original work"
AI isn't that

PR #775, opened 2026-05-06
Add ability to put physics constraints on any PhysicsPipelineBody

TheLMiffy1111:
Allow `PhysicsPipelineBody` as arguments of `addConstraint`.

The motivation is to allow addons to be able to add physically constrained rigid-bodies through `BoxPhysicsObject` instead of sub-levels as `BoxPhysicsObject` is easier to manage.

Since instances of `BoxPhysicsObject` aren't added to the set of level colliders, this requires getting the local center of mass of rigid-bodies to also default to the zero position vector if they are not found in the set.


ryanhcode:
This should now be supported in base sable

TheLMiffy1111:
Obsoleted by https://github.com/ryanhcode/sable/commit/9736e55

Issue #200, opened 2026-04-19
Spawnpoint persists after bed or respawn anchor destroyed


spekir-hash:
### Description

Destroying a bed after having set your spawn on a physics object does not count as spawn obstructed/bed destroyed. It still acts as your spawnpoint despite the fact no bed is present. This only changes once the spawnpoint is reset with another bed. Also works with respawn anchors in the nether.

### Game Log

sorry im too lazy, its a consistent bug. replicate it yourself pls.

Ocelot5836:
There is currently not an API for querying blocks in unloaded sub-levels. Works as intended for now

PR #812, opened 2026-05-08
Partially validate respawn for loaded/deleted sublevels

Syntax3rr:
Partial fix for #200 

If the sublevel is loaded or deleted, we try to check for beds/respawn anchors.
Unloaded sublevels are out of scope of this PR.

This partial solution means players aren't stuck in infinite death loops if say their airship falls into the void in dimensions without the void sea.


Ocelot5836:
This will be properly fixed when support for loading sub-levels is added

PR #830, opened 2026-05-09
Add necessary arguments to JNI call to allow destructive block callbacks

cakeGit:
On draft due to `Caused by: java.lang.RuntimeException: Rapier native panic: Internal error: island should be awake.` issues which apparently are elsewhere, this just really seems to cause it.


ryanhcode:
can we tag the overloads for backwards compatibility as deprecated?

ryanhcode:
!!

ryanhcode (approved):

ryanhcode (changes requested):
please remove the sable_rapier_binaries.zip.l4z change from this PR

ryanhcode:
this shouldn't be necessary after 9736e551c5d6da867e687dc21e026ed7bf9211e4, let me know if more changes are needed

cakeGit:
Nope that's perfect

PR #884, opened 2026-05-14
Add musl-linked Rapier natives for Alpine Linux

### LINKED ISSUE #13: Natives missing for musl-based distros
### Context

The `sable_rapier_natives` library doesn't support musl-based distributions like Alpine Linux (here using the `itzg/minecraft-server:java21-alpine` docker image).

ldd output:
```
$ docker exec -ti minecraft-1 ldd /tmp/sable_rapier_natives11103248724679144649.tmp
        /lib/ld-musl-x86_64.so.1 (0x7f9fd46e0000)
        libm.so.6 => /lib/ld-musl-x86_64.so.1 (0x7f9fd46e0000)
        libc.so.6 => /lib/ld-musl-x86_64.so.1 (0x7f9fd46e0000)
Error loading shared library ld-linux-x86-64.so.2: No such file or directory (needed by /tmp/sable_rapier_natives11103248724679144649.tmp)
        libpthread.so.0 => /lib/ld-musl-x86_64.so.1 (0x7f9fd46e0000)
        libdl.so.2 => /lib/ld-musl-x86_64.so.1 (0x7f9fd46e0000)
```

### Crash Report

https://mclo.gs/xQctQOJ

Botelho31:
Closes #13.

On musl-based Linux (e.g. Alpine in Docker), the server reported:

Sable has failed to load the natives needed for its Rapier pipeline. Native library name sable_rapier_x86_64_linux.so.
Linux artifacts were only built for *-unknown-linux-gnu (glibc). Alpine’s dynamic linker and libc are musl, so the existing *_linux.so binaries are not loadable there even though the architecture is still x86_64.

While debugging the same environment, this also appeared in the log:

[common.asm.RuntimeDistCleaner/DISTXFORM]: Attempted to load class net/minecraft/client/multiplayer/ClientLevel for invalid dist DEDICATED_SERVER
That message comes from loading a client-only class on a dedicated server (distribution / classloading). It is not fixed by shipping musl natives; if it persists after this change, it should be tracked and fixed separately.

Solution:

- Cross-compile Rapier JNI for x86_64-unknown-linux-musl and aarch64-unknown-linux-musl, package them as sable_rapier_<arch>_linux_musl.so, and include them in sable_rapier_binaries.zip.l4z alongside the existing glibc *_linux.so builds (common/build.gradle).
- Install the musl Rust targets in the zigbuild Docker image (common/src/main/rust/container/zigbuild/Dockerfile).
- Set -C target-feature=-crt-static for musl targets so the JNI cdylib links dynamically against libc (required for System.load JNI libraries) (common/src/main/rust/.cargo/config.toml).
- At runtime, detect musl and load …_linux_musl.so; otherwise keep loading …_linux.so (Rapier3D.java).

Note:
Maintainers must rebuild the zigbuild image and run common:buildRustNatives (or CI equivalent) so published jars contain the new zip entries; old builds without *_linux_musl.so will still fail on Alpine.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,2 +1,3 @@
+ - Ship musl-linked Rapier natives for Alpine Linux (`*_linux_musl.so`) alongside glibc builds
  - Fix issues with glibc support on Linux Sable Rapier natives
  - Fix Create Blueprints not working on sub-levels
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -22,6 +22,10 @@ def linux(arch) {
     [triple: "${arch}-unknown-linux-gnu", suffix: ".2.17", arch: arch, os: "linux", ext: "so", lib: true]
 }
 
+def linuxMusl(arch) {
+    [triple: "${arch}-unknown-linux-musl", suffix: "", arch: arch, os: "linux_musl", ext: "so", lib: true]
+}
+
 def freebsd(arch) {
     [triple: "${arch}-unknown-freebsd", suffix: "", arch: arch, os: "freebsd", ext: "so", lib: true]
 }
@@ -35,6 +39,8 @@ def supportedTargets = [
         mac("aarch64"),
         linux("x86_64"),
         linux("aarch64"),
+        linuxMusl("x86_64"),
+        linuxMusl("aarch64"),
         windows("x86_64"),
         windows("aarch64"),
 //        freebsd("x86_64"),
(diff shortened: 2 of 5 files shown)


IThundxr:
Thanks for the PR, i'm currently in the process of writing a gradle plugin to make shipping more natives easier and this'll be added alongside that.

PR #889, opened 2026-05-14
Some mixin compatibility improvements

invokekitty:
Looked at some mixins and made a few changes to improve compatibility with other mods. This also allows lithium's efficient raycast implementation to be used on sub levels, as `BlockGetterMixin` now wraps the method instead of overwriting it. 
Besides this i replaced some Redirects that just wrap the original call with WrapOperation, ModifyArg or ModifyExpressionValue, and some Injects that cancel at RETURN with ModifyReturnValue.

Please tell me if i should change anything :3


Ocelot5836 (approved):
LGTM

lap2ka:
For me this makes inspect mode in ponders unusable

Ocelot5836:
After talking with ryan it would be better if smaller prs to mixins were submitted with specific fixes for certain mods. It reduces the chance of a weird regression without any benefit

PR #891, opened 2026-05-14
Make rust building tasks configuration cache friendly

tmvkrpxl0:
This PR adds dedicate Task and switches from using `doLast`. This is to make it configuration cache friendly.
I did not enable configuration cache in gradle.properties because I'm not sure if there is any other incompatible task.
But in my testing I could launch Minecraft and build rust natives with no problem when it was enabled.

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -1,3 +1,11 @@
 plugins {
     id 'groovy-gradle-plugin'
 }
+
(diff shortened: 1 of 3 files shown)


IThundxr:
Thanks for the PR, however this code is already being rewritten at the moment to more properly interact with gradle and its caching.

PR #901, opened 2026-05-15
Fix Breeze charges not affecting contraptions

### LINKED ISSUE #618: Breeze charges and wind charging debuff do not affect contraptions
### Description

Breeze wind charges and the wind charging debuff do not push contraptions in any way, which is a disparity when compared to regular wind charges which push contraption as one would expect.
To recreate, simply kill a mob with the potion effect or get a breeze to shoot a single block contraption. The block will not move at all, unlike with a wind charge from a dispenser, etc.

### Game Log

https://mclo.gs/budQzAP

EZForever:
Breeze charges are instanceof `BreezeWindCharge`, not `WindCharge`. By checking their common super class, `AbstractWindCharge`, we can support both kinds of wind charges to affect contraptions.

Partially fixes #618.


ryanhcode:
thank you!

PR #905, opened 2026-05-16
Fix nether portal linking (#793): don't scale Y by dimension multiplier

### LINKED ISSUE #793: Netherportal Linking Behavior Affected
### Description

Possible Nether portal linking issue with Sable enabled

I’m seeing behavior where vanilla Nether portal linking appears to choose the higher Overworld portal even when a lower portal is closer by Euclidean distance.

Example setup:

- Source Nether portal: (-288, 118, -48)
- Higher Overworld portal: (-2304, 194, -395)
- Lower Overworld portal: (-2304, 193, -384)

When traveling from the Nether portal at (-288, 118, -48), the game links to the higher Overworld portal at approximately (-2304, 194, -395), even though the lower Overworld portal at approximately (-2304, 193, -384) should be closer.

Calculated distances from the expected Overworld target:

- To higher Overworld portal: 77.47 blocks
- To lower Overworld portal: 75.11 blocks

Expected behavior:
Vanilla portal linking should choose the closer portal, which should be the lower Overworld portal.

Actual behavior:
With Sable enabled, the Nether portal links to the higher Overworld portal instead.

Troubleshooting performed:
- Breaking/disabling the higher Overworld portal allows the lower Overworld portal to work correctly.
- Re-enabling the higher Overworld portal causes the link to prefer the higher portal again.
- Disabling Sable appears to restore expected vanilla portal-linking behavior.

This makes it seem like Sable may be affecting portal POI selection, portal distance calculation, or the way candidate portals are prioritized.

ModLoader: NeoForge
Sable Version: 1.2.2
Single Player

### Game Log

https://mclo.gs/uHGNxZm

QueenRose4444:
Fixes #793

Problem
NetherPortalBlockMixin applies the dimension coordinate multiplier (8x for Nether→Overworld, 0.125x for Overworld→Nether) to all three axes (X, Y, Z) when calculating the portal destination target position. Vanilla Minecraft only scales X and Z — the Y coordinate is never scaled by the dimension multiplier.

This causes the portal search target to have a wildly incorrect Y value (e.g., Y=118 becomes Y=944 when going Nether→Overworld), resulting in the game selecting the wrong destination portal.

Fix
Remove * multiplier from the Y coordinate in the clampToBounds call (line 31 of NetherPortalBlockMixin.java).

Tested
Verified fix resolves the issue in both singleplayer and multiplayer. Portal linking now correctly selects the closest portal by Euclidean distance, matching vanilla behavior.

### DIFF
--- a/common/src/main/java/dev/ryanhcode/sable/mixin/portal/NetherPortalBlockMixin.java
+++ b/common/src/main/java/dev/ryanhcode/sable/mixin/portal/NetherPortalBlockMixin.java
@@ -28,7 +28,7 @@ public class NetherPortalBlockMixin {
 
         return instance.clampToBounds(
                 globalPos.x * multiplier,
-                globalPos.y * multiplier,
+                globalPos.y,
                 globalPos.z * multiplier
         );
     }


ryanhcode:
thank you!

PR #415, opened 2026-04-22
Add light propagation between sub-levels, world

izekblz:
# Overview
Implements bidirectional light propagation across all boundaries: world ↔ sub-level, sub-level ↔ world and sub-level ↔ sub-level. Also fixes compatibility with dynamic lighting mods and adds backward compatibility shims for dependent mods (I believe this was already fixed? Although still with Override usage. This is singled out in a separate commit anyways)

Those features are by no means ready-to-ship and are better viewed as an experimental build. However, after testing i did not find any major visual bugs present.

# Light Propagation
## World → Sub-level (server-side):

`ServerSubLevelLightInjector` scans world blocks near each sub-level's visual bounding box for light emitters

Transforms world positions into plot-local coordinates and injects emission values directly into the plot's `BlockLightEngine` via `setStoredLevel` + `enqueueIncrease`

Rescans on world light changes, sub-level creation/split, and every 8 blocks of movement (so the brightest light still isn't missed, with brightest being 16 i believe and 8 being half of that)

Per-source NPE handling for plot sections without allocated DataLayers

## Sub-level → World (client-side):

`VirtualLightManager` collects emitting blocks from sub-level plots, transforms to world space, and spreads virtual light with manhattan distance falloff

Injected into vanilla via `BlockLightSectionStorage` mixin and Sodium via `LevelSlice` mixin

Server-side implementation would have to work with `ThreadedLevelLightEngine` and i don't have a clue how to do that

## Sub-level → Sub-level (server-side):

`fullRescan` iterates nearby sub-levels' plot chunks, reads block states for emitters, transforms positions to world space via `Pose3d.transformPosition`, and injects into the receiving sub-level's plot light engine

`LevelChunkMixin` detects when a block's light emission changes on a sub-level plot (comparing old vs new state) and marks nearby sub-levels for rescan

Moving sub-levels mark nearby sub-levels dirty so they pick up emitters at updated positions

Old injected positions cleared unconditionally during reinject (plot-local coords can't be validated against world state)

# Performance
Light updates sent via `ClientboundLightUpdatePacket`, no full chunk resends and avoiding client-side state resets

# Dynamic Light Compatibility
EntityRendererMixin: Changed @Overwrite to @Inject + @Redirect for `getPackedLightCoords` to preserve other mods' injections into `getBlockLightLevel`

[...]

PR #916, opened 2026-05-16
feature/light propagation

izekblz:
# Intro

Another attempt of mine to add light propagation to\from sublevels, now fully done on server side, so injected light also has gameplay effects instead of being client side visual

There are a lot of architectural changes since the last PR, also did a pass over that version to comply with SLA

Current version fully working and can be tested if you wish to check performance, however I fully understand that such changes shouldn't be just merged into main. My goal is to propose the architecture and share my draft for your consideration. Edge cases i haven't fixed will be listed in the end

I have been hosting a server with this updated version for a week and so far hasn't recieved any complaints from players and didn't notice much issues myself. Performance difference between main Sable and this fork is almost negligible, if Spark is to be trusted

# Architecture

There are two light injectors (world->plot and reverse) and one opacity mixin

## `ServerSubLevelLightInjector`, world->plot

Runs from ServerLevelPlot.tick() per sub-level. Injects world emitters and emitters from other sublevels (that are injected to world).

`tickPlot` detects movement and does either:
- `fullRescan` -- mostly used on the first tick to get initial state, scans full light box (16 blocks in each direction)
- `incrementalScan` -- shift the cached emitter set and scan only the leading slice(s), done on translation
- `boundsDiffScan` -- scan only regions added by rotation

`scanOtherSubLevelEmitters` finds emitters on overlapping sub-levels and projects them in

`scanWorldOpaqueIntoPlot` projects world and other sub-levels opaque blocks into plot space, stored in plotLocalWorldOpaque

`reinject` clears old positions in the plot's LevelLightEngine, writes the new ones via setStoredLevel + enqueueIncrease, and runs propagation

## `ServerSubLevelWorldInjector`, plot → world

Runs from ServerSubLevelContainer.tick() once per server tick, after every sub-level's tickPlot has run. Has five phases:

1. `detectMovement` -- edge crossing on bounding box corners flags the sub-level for update
2. `rebuildPlotLocalCache` (only on block change) reads the plot once and stores a `PlotLocalLightData` record: arrays of plot positions for emitters, opaque blocks and shape-occluding blocks (slabs/stairs). `projectToWorldSpace` then transforms those cached positions into world space using the current pose
3. The merged opaquePositions, `opaquePositionsCore` (no gap-fills) and `shapeOcclusionMap` are rebuilt from per sub-level slices, world emitters near changed sub-levels are then re-propagated on the light thread (clear + re-emit), so they respect the new opacity geometry
4. Sub-levels whose emitters overlap the changed bounds get marked for reinjection
5. Each changed sub-level's emitters are pushed into the world `LevelLightEngine` via `taskMailbox`

## Caches

Main optimization since last version i submitted -- sub-levels now cache a lot of stuff to reduce full scans

| Cache | Description | Rebuild time |
| ----------- | ----------- | ----------- |
| `PlotLocalLightData` per sub-level | Plot emitter / opaque / shape arrays | Block change in plot |
| `cachedPlotSources` per sub-level | Last world-space emitter projection | Sub-level moves or rotates |
| `perSubLevelOpaque`, `perSubLevelOpaqueCore`, `perSubLevelShapes`| Per-sub-level opaque & shape projections; *core* excludes diagonal-leak gap-fills so sub-level doesn't shadow itself | Sub-level moves or rotates |
|`opaquePositions` / `opaquePositionsCore` / `shapeOcclusionMap` | Same but global merge | Any perSubLevel change |
| `cachedWorldSources` (per sub-level) | Last world-emitter scan, so incremental and bounds-diff scans only have to read the delta | Sub-level moves or rotates |
| `plotLocalWorldOpaque` (per sub-level) | World and other sub-levels opaque, projected into plot | On move or world block change near the sub-level |
| `injectedPositions` (per sub-level) | Where this sub-level contribution currently lives in the light engine, so we can clear it before re-injecting | Each tick |

# Known bugs

- Single block sub-levels don't get lit correctly -- idk why
- Fully blocking stairs or slabs don't actually block propagation -- would require expensive voxel-shape check, can be done but unsure if worth it
- Sometimes light source located on sub-level still illuminates scene when fully inside world block -- i have a check for source-in-block but it just doesn't trigger sometimes idk, rare case anyway


common/src/main/java/dev/ryanhcode/sable/render/light_bridge/ServerSubLevelWorldInjector.java
```diff
@@ -0,0 +1,641 @@
...
+ * Server-side light injection from sub-level plots into the world (plot → world direction).
+ * <p>
+ * Handles three unified concerns triggered by sub-level movement or block changes:
+ * <ul>
+ *   <li>Emitter projection: sub-level light sources → world light engine</li>
+ *   <li>Opacity projection: sub-level opaque blocks → light propagation barrier</li>
+ *   <li>Neighbor notification: when a sub-level changes, nearby sub-levels re-propagate</li>
+ * </ul>
+ */
+public final class ServerSubLevelWorldInjector {
```
cakeGit, line 42: The class is named quite vaguely and doesen't mention light, i would call it something like "SubLevelToWorldLighter" and the other class be called "WorldToSubLevelLighter"

common/src/main/java/dev/ryanhcode/sable/render/light_bridge/ServerSubLevelWorldInjector.java
```diff
@@ -0,0 +1,641 @@
...
+
+    private static final double LIGHT_MARGIN = 15.0;
+
+    private static final ConcurrentHashMap<UUID, LongSet> injectedPositions = new ConcurrentHashMap<>();
```
cakeGit, line 46: This many ConcurrentHashMaps seems unnecassary, could you not hold a single ConcurrentHashMap<UUID, SubLevelLightInfo> or something

common/src/main/java/dev/ryanhcode/sable/render/light_bridge/ServerSubLevelWorldInjector.java
```diff
@@ -0,0 +1,641 @@
...
+    public static void onPlotBlockChanged(final ServerSubLevel subLevel) {
+        final UUID id = subLevel.getUniqueId();
+        needsPlotRescan.put(id, Boolean.TRUE);
+        needsUpdate.put(id, Boolean.TRUE);
+    }
+
+    /**
+     * Called once per server tick after all plots have ticked, from ServerSubLevelContainer.tick().
+     */
+    public static void tick(final ServerLevel level) {
```
cakeGit, line 133: Incredibly long tick needs to be broken down,
I call it comment blocking but each of the phases seem they could probably be much better as separate methods

common/src/main/java/dev/ryanhcode/sable/render/light_bridge/ServerSubLevelLightInjector.java
```diff
@@ -0,0 +1,642 @@
...
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Server-side light injection between sub-levels and the world (world → plot direction).
+ * Scans world emitters near sub-levels and injects them into the plot's light engine.
+ */
+public final class ServerSubLevelLightInjector {
+
+    private static final double LIGHT_MARGIN = 15.0;
```
cakeGit, line 35: Repeated constant

common/src/main/java/dev/ryanhcode/sable/physics/impl/rapier/RapierPhysicsPipeline.java
```diff
@@ -458,6 +458,7 @@ public void handleChunkSectionAddition(final LevelChunkSection section, final in
         // if it's only air, all zeros will do. it'll default to empty neighborhood state and 0 (empty) collider ID
         if (!section.hasOnlyAir()) {
             final LevelChunk chunk = this.accelerator.getChunk(x, z);
+            if (chunk == null) return;
```
cakeGit, line 461: Whats the reason for this? As far as i can see accelerator.getChunk is nonnull

common/src/main/java/dev/ryanhcode/sable/render/light_bridge/PlotLocalLightData.java
```diff
@@ -0,0 +1,33 @@
+package dev.ryanhcode.sable.render.light_bridge;
+
+/**
+ * Cached plot-local block data for a sub-level's light-relevant blocks.
+ * Populated once on block change, re-projected to world space on movement.
+ */
+public final class PlotLocalLightData {
```
cakeGit, line 7: This could be a record

common/src/main/java/dev/ryanhcode/sable/render/light_bridge/ServerSubLevelLightInjector.java
```diff
@@ -0,0 +1,642 @@
...
+                needsReinject.put(id, Boolean.TRUE);
+            }
+        }
+    }
+
+    /**
+     * Called when a light-emitting block changes on a sub-level's plot.
+     */
+    public static void onPlotBlockLightChanged(final ServerLevel level, final ServerSubLevel changedSubLevel) {
+        final var container = dev.ryanhcode.sable.api.sublevel.SubLevelContainer.getContainer(level);
```
cakeGit, line 98: Theres consistent use of fully qualified names instead of imports

cakeGit (commented):
Im not an authority here but im concerned about the quality of this code. From my experience, the fully qualified names, duplicated code and overly complex structure indicates this is written with some coding agent without enough oversight, just like the last PR. This is a very large system and should be much more carefully implemented, by someone who understands the system better. This should be closed the same as #415

common/src/main/java/dev/ryanhcode/sable/physics/impl/rapier/RapierPhysicsPipeline.java
izekblz: For some reason got crashes on my server on world load, logs suggested possible chunk load before it initialized (even with base Sable build), didn't dig too deep there. Although I was supposed to remove that before doing a PR, whoops

common/src/main/java/dev/ryanhcode/sable/render/light_bridge/ServerSubLevelWorldInjector.java
izekblz: Yeah that would also make cleanup easier, although there's a possible race on injectedPositions and that maybe should've been volatile and separated somehow? Or just make this all into a record idk

common/src/main/java/dev/ryanhcode/sable/render/light_bridge/ServerSubLevelWorldInjector.java
izekblz: Yeah that was once a simple detect movement - rescan - reproject and was readable, at this point it's too loaded

izekblz:
Thanks for checking this out and leaving meaningful comments! As I've noted in the PR, it's mostly a showing of one possible way to do light -- I did not intend for this to be used in a production environment. Hope you guys will at least get an idea or two out of this for the feature :)

PR #929, opened 2026-05-17
Fix standing up from crawling

### LINKED ISSUE #7: Unshifting while crouched causes the player to "clip into" the slab
### Description

Unshifting while crouched causes the player to "clip into" the slab, in worldspace the player cannot unshift while under a slab

Image

Image

### Game Log

no logs produced

benjaminiserman:
I'm going to deploy this fork to a private server for testing, but this fixed the issue in my initial testing for both:
- when entering crawling via a trapdoor on the main level, then crawling under a sable sublevel
- when entering crawling via a trapdoor while on a sable sublevel

### DIFF
--- a/common/src/main/java/dev/ryanhcode/sable/mixin/player_standup/PlayerMixin.java
+++ b/common/src/main/java/dev/ryanhcode/sable/mixin/player_standup/PlayerMixin.java
@@ -1,0 +1,32 @@
+package dev.ryanhcode.sable.mixin.player_standup;
+
+import dev.ryanhcode.sable.mixinhelpers.CanFallAtleastHelper;
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.level.Level;
+import net.minecraft.world.phys.AABB;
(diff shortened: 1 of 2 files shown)


Ocelot5836 (approved):
LGTM, will wait for further testing

benjaminiserman:
Planning to deploy to my test server late tonight. I have a big airship with 1-block tall maintenance corridors that will be perfect for testing this. Have fallen through the floor a lot of times on the current version lol

benjaminiserman:
So far so good with the testing, though it looks like we'll need an additional mixin to handle standing up from Create's cardboard armor. Not a blocker for this PR though.

benjaminiserman:
Note from further testing: this change does tend to cause the player to crawl a lot when near the walls of a ship that's rotating. This is still better than players getting shunted through the floor, but it might be worth building a more advanced system for relaxing the player hitbox when sublevels aren't Y-axially aligned.

Easyidle123:
It seems this change is responsible for being forced to crawl when the airship is moving upwards quickly as well, which is quite intrusive to gameplay

PR #939, opened 2026-05-18
Moved Plotyard back to being far beyond the natural world border

Hectoris919:
Restored the commented out code for DEFAULT_ORIGIN.

Having DEFAULT_ORIGIN = 10000 is a safe number for normal players, but there are edge cases such as anarchy servers where crazy and dedicated players may attempt to travel to the plotyard and grief other player's ships, or even attempt to fly a ship into its own plot, which could cause undesired results.

While this could be resolved by shrinking the world border, it is entirely possible for a survival player to visit the plotyard on their own if they wanted to.


Cyvack:
This really is not a valid solution to the issue. It may do you good to do proper testing to understand why the plotyard is not outside of the world border by default, and why that value is commented out.
The actual solution will take more than just un-commenting out a single line.

Issue #170, opened 2026-04-19
Item Duplication


SAILQAQ:
### Description

When vanilla or modded containers are assembled, the items stored inside them get duplicated.
For details, see: https://github.com/Creators-of-Aeronautics/Simulated-Project/issues/107, https://github.com/Creators-of-Aeronautics/Simulated-Project/issues/155, https://github.com/Creators-of-Aeronautics/Simulated-Project/issues/208, etc. 

### Game Log

Please consult the attachments on the linked issues for the relevant debugging data.

SAILQAQ:
Farmer's Delight's cooking pot has the same issue (including pots from all add-on mods as well). 

<img width="2560" height="1494" alt="Image" src="https://github.com/user-attachments/assets/49d11f09-ae47-4401-8117-1c8d2601ebb7" />

PR #951, opened 2026-05-19
Fixed a bug that duplicated items. 

yansunsky:
The bug was: when an item was placed on a Table Cloth, using the Physics Assembler would cause the item to be duplicated. The fix was based on the fact that blocks like Table Cloth are not registered as clearable, so an additional method was written to handle blocks not registered as clearable. The implementation approach is to directly initialize the block's NBT, retaining only its own ID, since it will be removed soon.
This is my first time using a pull request. I'm not sure if my fix is a good solution, but it should be useful.


yansunsky:
Uploading 202605192308_20260519_23104878.mp4…

yansunsky:
fix #170

Ocelot5836:
Fixed by https://github.com/Creators-of-Create/Create/pull/10400

PR #952, opened 2026-05-19
Fix minecarts disappearing when derailing on sublevel

JpegN:
`AbstractMinecartMixin`: when a cart isn't on rails anymore (`!isOnRails()`), kick it back to world space right away instead of waiting for it to leave plot bounds and killing it. Server-only.

`ClientPacketListenerMixin.sable$lerp`: the sub-level to world branch was using `setPos`, which doesn't touch `xo/yo/zo`. That caused a one-frame flicker since rendering would interpolate between shipyard and world coords, drawing the entity at a ~15M block midpoint. Swapped it for `moveTo`. Fixes the flicker for any entity, not just minecarts.

Minecart entries in `destroy_when_leaving_plot.json` aren't used anymore.

### DIFF
--- a/common/src/main/java/dev/ryanhcode/sable/mixin/entity/entities_stick_sublevels/ClientPacketListenerMixin.java
+++ b/common/src/main/java/dev/ryanhcode/sable/mixin/entity/entities_stick_sublevels/ClientPacketListenerMixin.java
@@ -105,7 +105,8 @@ public abstract class ClientPacketListenerMixin {
             if (subLevel != null && actuallyInSubLevel && existingSubLevel != subLevel) {
                 entity.setPos(subLevel.logicalPose().transformPositionInverse(entity.position()));
             } else if (existingSubLevel != null && subLevel == null) {
-                entity.setPos(existingSubLevel.logicalPose().transformPosition(entity.position()));
+                final Vec3 newPos = existingSubLevel.logicalPose().transformPosition(entity.position());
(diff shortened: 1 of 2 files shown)


Derpgamer22:
I was looking through here and I saw this and personally, I think this is amazing. Curious, do you think something like this is possible for armor stands as well? It'd be nice so if for whatever reason your armor stand falls off a ship, it doesn't just dematerialize with possible armor on it that would vanish with it :p

I do get that this is a weird spot to ask something like this... so sorry if me putting this comment here is something I shouldn't be doing.

JpegN:
> I was looking through here and I saw this and personally, I think this is amazing. Curious, do you think something like this is possible for armor stands as well? It'd be nice so if for whatever reason your armor stand falls off a ship, it doesn't just dematerialize with possible armor on it that would vanish with it :p
> 
> I do get that this is a weird spot to ask something like this... so sorry if me putting this comment here is something I shouldn't be doing.

Possible, yes. But not by the same logic, the minecarts can simply go by isOnRails() or !isOnRails(). needless to say, armor stands don't use rails. 
The obvious solution would be to check if the armor stand is grounded, then kicking it off the sublevel if not.

If you have any experience with java, or want to learn, give it a try! I'll make my own fix if needed when I'm next available (probably Sunday).

JpegN:
apologies, renamed the branch thus auto-closing the pr by accident.

PR #960, opened 2026-05-19
Shoulder Surfing Compat

### LINKED ISSUE #913: Shoulder surfing reloaded
### Context

This (should be) the crash log right after switching to a front facing third person view with both aeronautics and shoulder surfing reloaded installed. I'm pretty sure this is where I submit compatibility issue reports 

### Crash Report

https://mclo.gs/H9GFPfO

Ocelot5836:
* Implements compat for shoulder surfing
* The shoulder surfing camera angle is incorrect when riding on a sub-level


EarthKing339:
How do I use or implement this into everything

YassiGame:
@EarthKing339 
> How do I use or implement this into everything

Just keep in mind that this is a draft PR, so it may not work yet (but the build passes so maybe everything is fine)

You will need git and java 21 (or any higher... i think)
```BASH
# Open the terminal in a empty folder

git clone https://github.com/ryanhcode/sable.git .
git fetch origin pull/960/head:pr-960
git checkout pr-960

# MacOS or Linux
./gradlew build

# Windows
gradlew.bat build
```

After that you wil have a built jar of sable with this fixes in neoforge/build/libs (I’m saying this from memory, I’m a bit too lazy to check the source code, but that should be the right path)

Exopandora:
> The shoulder surfing camera angle is incorrect when riding on a sub-level

The root cause for that is that Shoulder Surfing calls `Camera.setRotation` once again after `Camera.move` for camera sway (see [MixinCamera](https://github.com/Exopandora/ShoulderSurfing/blob/master/common/src/main/java/com/github/exopandora/shouldersurfing/mixins/MixinCamera.java#L100) of Shoulder Surfing). The mixins of sable then apply the sublevel rotation an additional time (see [CameraMixin](https://github.com/ryanhcode/sable/blob/main/neoforge/src/main/java/dev/ryanhcode/sable/neoforge/mixin/camera_rotation/CameraMixin.java#L90) of sable), which messes up the camera position. A fix would be to not change the implementation of `Camera.setRotation`, but instead use a redirect or to apply the camera rotation adjustments directly after each `setRotation` call in the `setup` method.

Issue #938, opened 2026-05-18
Vertical build limits


tatoyoda600:
### Description

When creating a Sable sublevel, the first block is placed at Y=128 (within the sublevel) in the current dimension, this means that the upwards and downwards build limit for the sublevel are:
- Overworld: 192 upwards, 192 downwards
- Nether: 128 upwards, 128 downwards
- The End: 128 upwards, 128 downwards

If you want to build a 200 block tall structure, you need to make sure to first go at least 8 blocks down (72 in the Nether or End) from your initial block, because otherwise you'll hit the build limit after 192 blocks (128 in the Nether or End).
In theory any dimension with a height limit lower than 128 will cause issues (Potentially related to https://github.com/ryanhcode/sable/issues/229?)

Personally, I was just trying to make a blimp with a drill using Create: Simulated, but my drill was randomly stopping around Y=100 for seemingly no reason. I checked my configs for the max Rope Pulley length and this was way under the limit. Only after a long while of trying things, seeing how different blimp designs I made reached different depths, and trying to find some sort of logic, did I eventually notice that the tip of my drill had a Y position of -64 in F3 (within the sublevel of course).
This means that I now need to move all of my blimp's blocks up a couple hundred blocks, within the same sublevel, in order to actually have a usable drill.

<img width="1747" height="986" alt="Image" src="https://github.com/user-attachments/assets/82a91eb6-db10-4693-a816-31974089f250" />

### Suggestions:
1. Ideally the sublevel would shift, grow, or adjust in some way to allow the player to use the full sublevel height without needing to plan ahead, but I can imagine that wouldn't be easy, and might not be possible.
- With this, my drill would automatically make space for itself as it lowers, and would only hit a limit when the full sublevel height is reached

2. Alternatively, having sublevels start at the actual Y level that they were started on could be a good compromise. This would mean that if you're in the Overworld and you create a sublevel near bedrock, you'll be allowed to build up to the height limit (around 300 blocks up), but wouldn't be allowed to extend the sublevel much downwards (as the initial block would've been placed at around Y=-64).
- This would allow me to build my blimp directly at Y=300, without needing to spend time extending the sublevel vertically so that the sublevel has downwards space available

[...]

PR #980, opened 2026-05-21
Configurable sub-level assembly height (#938)

tatoyoda600:
Adds 2 config values, implementing suggestions 2 & 3 from #938:
- SUB_LEVEL_FIXED_START_HEIGHT: If SUB_LEVEL_START_POS is set to FIXED, then this is the height (Default value of 128 matches the current approach)
- SUB_LEVEL_START_POS: Determines the approach to use for determining the height (Y value) of the first block in a newly assembled sub-level (FIXED, DYNAMIC, MIDPOINT)
  - FIXED: Uses the height specified by SUB_LEVEL_FIXED_START_HEIGHT (Essentially the current approach)
  - DYNAMIC: Uses the world height of the position the block is in (Spawn a sub-level at Y=73, the block is placed at Y=73 in the sub-level)
  - MIDPOINT: Calculates the midpoint between the min and max build limits, then uses that as the height

Defaults are 128 and FIXED, to match the current approach exactly, with all the customization being exclusively opt-in.

----

With dynamic assembly height, a 1-block sub-level created at Y=87 has its sole block located at Y=87 within the sub-level
image

### DIFF
--- a/common/src/main/java/dev/ryanhcode/sable/SableConfig.java
+++ b/common/src/main/java/dev/ryanhcode/sable/SableConfig.java
@@ -3,6 +3,7 @@ package dev.ryanhcode.sable;
 import net.neoforged.neoforge.common.ModConfigSpec;
 
 public final class SableConfig {
+    public enum SubLevelStartPos { FIXED, DYNAMIC, MIDPOINT }
 
     public static final ModConfigSpec SPEC;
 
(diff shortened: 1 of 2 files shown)


tatoyoda600:
Implemented suggestion 1 from #938, but it kinda sucks, so not including in this PR (https://github.com/tatoyoda600/sable/tree/automatic-height-adjustment)
It has some uses and would be great if it could be implemented well, but I don't know how feasible (or worth it) it would be to actually polish it up to a point where it would be truly usable.

---

It does help with the issue I was having in #938 though, as soon as the Rope Pulley reaches -64 in the sub-level, the sub-level's contents are shifted up to give it more downwards space to expand, letting my Rope Pulley reach real deep.
image
image

PR #981, opened 2026-05-21
Fix block duplication on assembly

Ocelot5836:
Allows some blocks to be overridden when they do no implement Clearable correctly with the `sable:silent_assembly_removal` tag.

Also implements a manual game test to check all block states in the registry for correct state assembly if they have an item stack capability.


ryanhcode (approved):
looks swag to me!

Issue #877, opened 2026-05-13
Issue with Armor stands leaving sublevels


Derpgamer22:
So, I am aware that is some entity tags that Sable has that does some things. One of these tags causes an entity to delete if it falls out of a sublevel, and the armor stand has that tag.

Apparently, the armor stand, when deleted, doesn't drop anything, being itself and anything that is on the armor stand, like armor...

I am pretty sure the issue could be solved if the armor stands didn't have that tag... I mean... why do they have to delete when they fall off of a sublevel anyway? I know that they fill fall through the ground...

It would be nice if when they fall out of the sublevel, that they would orientate into the actual world so they could place... same goes for minecarts.

By the way, this tag I mentioned also causes issues with supplementaries' spring launcher, as when an entity is launched from it and it goes out of the sublevel, the entity disappears. I was able to test an entity that doesn't have this tag, which being the Command Block in a Minecart, since it doesn't actually have the tag, and it allows it to function completely fine.

Derpgamer22:
I also learned that you cannot push minecarts that are connected to a sublevel.

Derpgamer22:
https://cdn.discordapp.com/attachments/1230534553618350144/1504180759492169829/Screen_Recording_2026-05-13_120210.mp4?ex=6a060cff&is=6a04bb7f&hm=e92ecd9a4030266d6764df77b417d987d0547acbdf8e16e80808de2ab78c3f29&

here is a video of some example stuff

Derpgamer22:
I actually found out that the armor stand CAN survive and place back into the world if it is on a single sublevel block

Derpgamer22:
https://cdn.discordapp.com/attachments/1230534553618350144/1504181497522028684/20260513-1759-05.0835772.mp4?ex=6a060daf&is=6a04bc2f&hm=0387ee167ba0a6a527daca6e47a729794af2c399fe5b4b40dd07ce199f4b5ef3&

I'm sorry that the videos have to be downloaded... it keeps saying "failed to upload" :T

Derpgamer22:
maybe if it could be done where instead of entities getting destroyed when falling out of a sublevel, they just get moved back to the normal world plane

PR #982, opened 2026-05-21
Fix armor stands disappearing when falling off sub-levels

JpegN:
New `ArmorStandMixin` mirrors the minecart fix: when an armor stand is in a sub-level and not on the ground, kick it back to world space instead of letting it fall through shipyard and eventually get killed by the leave-plot tag. Server-only.

`destroy_when_leaving_plot.json` no longer needs the armor stand entry.

### DIFF
--- a/common/src/main/java/dev/ryanhcode/sable/mixin/entity/entities_stick_sublevels/ClientPacketListenerMixin.java
+++ b/common/src/main/java/dev/ryanhcode/sable/mixin/entity/entities_stick_sublevels/ClientPacketListenerMixin.java
@@ -103,9 +103,11 @@ public abstract class ClientPacketListenerMixin {
             final SubLevel existingSubLevel = Sable.HELPER.getContaining(this.level, entity.position());
 
             if (subLevel != null && actuallyInSubLevel && existingSubLevel != subLevel) {
-                entity.setPos(subLevel.logicalPose().transformPositionInverse(entity.position()));
+                final Vec3 newPos = subLevel.logicalPose().transformPositionInverse(entity.position());
+                entity.moveTo(newPos.x, newPos.y, newPos.z);
             } else if (existingSubLevel != null && subLevel == null) {
-                entity.setPos(existingSubLevel.logicalPose().transformPosition(entity.position()));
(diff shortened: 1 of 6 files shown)


JpegN:
I've found extra issues caused by my fix, will attempt to work them out.

JpegN:
Armor stands can now land on sublevels and actually 'attach' to them, this fixes *almost* all issues.
If someone does a lot of weird interactions with an armor stand going on and off a sublevel it could still fail and disappear, but for normal play this should be absolutely fine. (someone smarter than me could probably resolve it completely but I'm not that person)

Fixes issue #877.

Derpgamer22:
omg let's go :D

I do want to ask... would this work with supplementaries' hat stands? (The hat stand would have to have the tag that lets them retain on a sublevel would have to be given to them)

Also, since with this, armor stands would attach themselves to the sublevel when they fall on it... I wonder if it would be possible to have an armor stand on a sublevel with frost walker turn water into ice in the actual world. I had this machine thing that would have my sublevel create ice under it by using armor stands that were on the sublevel... but not at the same time...

JpegN:
> omg let's go :D
> 
> I do want to ask... would this work with supplementaries' hat stands? (The hat stand would have to have the tag that lets them retain on a sublevel would have to be given to them)
> 
> Also, since with this, armor stands would attach themselves to the sublevel when they fall on it... I wonder if it would be possible to have an armor stand on a sublevel with frost walker turn water into ice in the actual world. I had this machine thing that would have my sublevel create ice under it by using armor stands that were on the sublevel... but not at the same time...

This specific fix will not do anything for supplementaries' hat stand as is, or anything similar for that matter, I built it specifically on armor stands. Though it could be generalized to allow for easier compatibility, there were plenty of edge cases with the armor stand so I wouldn't be surprised to see even more with other similar entities.

As for the frostwalker thing, it will require a frostwalker-specific feature to be made to allow for the enchantment to work between sublevel and local world. That would be something for a feature request rather than a fix.

Derpgamer22:
Ah alright, well I do want to say that if this PR does get accepted, if there could be compat added for supplementaries' hat stand with this by the team

PR #986, opened 2026-05-21
Firing Chunk Loaded events on the server

boberchik342:
This fixes chunk loaded events not firing on the server.


boberchik342:
Should I do the same for unloading events?

PR #1042, opened 2026-05-26
Fix collision issues on sub-levels with dying entities

### LINKED ISSUE #593: Being Processed by Crushing Wheels Freezes the Server
### Context

RyanGPT, please write me a bug report that explains in more detail that when I jump into an assembled pair of crushing wheels and die from them, the entire server has a stroke, freezes, then stops me from respawning. Thank you

### Crash Report

https://mclo.gs/mR5wVbE

AdoLovesChicken:
When a player dies inside of a pair of Create's Crushing Wheels while standing on/in a sub-level, the server would either crash or delay the respawn of the dead player, causing them to wait in the respawn screen for quite a while. This is due to the velocity given to an entity colliding with a crushing wheel if on the ground by Create.

This PR should fix the issue where it would delay the player's respawn and it should return normal coordinates and not exceed any AABB limits.

Fixes #593


Ocelot5836:
This doesn't resolve the actual issue and instead just tries to add a patch over the symptoms

PR #1093, opened 2026-06-02
Fix crash caused by removing rapier rigidbody before chunk unload finished

WaterOtter86:
Sable was removing the Rapier rigid body before Minecraft finished unloading the sublevel chunk. If a block entity touched/updated or did anythin with blocks during its chunk unload callback, Sable’s blockchange hook tried to update mass/physics and somewhy teleport? data for a body that no longer existed in Rapier, crashing the server thread. Sorry for lib.rs touch, was testing stuff.


WaterOtter86:
stupid github-actions

PR #1120, opened 2026-06-05
Tagify block properties

Lucky-56:
so far `#sable:end_stone_slabs` instead of just having the `minecraft:end_stone_brick_slab`


Lucky-56:
should I do more like flywheel, grindstone and mud?

Squawkykaka:
Shouldn't this be using the "c" namespace, as if I'm remembering right that's the standard for forge based mods

Lucky-56:
also fabric, that's why it's `c` for common

Lucky-56:
`sable:end_stones` is used for the end stone properties tho

Ocelot5836:
This is a really niche use case and can be done if there's a need for it

PR #1136, opened 2026-06-07
Prevent lightning on sublevels

cakeGit:
Prevent lightning/skeleton traps from spawning inside the plot.
Since otherwise this just increases the density of lightning as you have more sublevels in an area, and these lightning strikes will just strike entities since they're the only real targets in the plot's lightning target checks.


Ocelot5836 (approved):
LGTM

Issue #1080, opened 2026-05-31
Client-side NPE in ClientboundSableUDPActivationPacket when UDP channel is null (tunnel/proxy environments)


blue-buff:
## Describe the bug

When playing behind an NAT traversal tunnel (e.g., EasyTier-based Terracotta/陶瓦联机), the client crashes with a `NullPointerException` in `ClientboundSableUDPActivationPacket.handle()` because the UDP `channel` is null. This causes the client to be kicked with "Invalid player data" (无效的玩家数据).

The issue is **client-side**, distinct from #175 (server-side NPE in `SableUDPServer.getServer()`).

## Steps to reproduce

1. Use any tunneling/proxy tool that creates a virtual network interface (EasyTier, frp, ngrok, etc.)
2. Join a server with Sable installed
3. The server sends `sable:udp_activation` → client's UDP channel is null → NPE

## Crash log

```
[Render thread/INFO] [dev.ryanhcode.sable.Sable/]: Client UDP channel inactive
[Render thread/INFO] [dev.ryanhcode.sable.Sable/]: Closed UDP channel!
[Render thread/INFO] [dev.ryanhcode.sable.Sable/]: Received authentication request, sending response over UDP to localhost/127.0.0.1:25565
[Render thread/ERROR] [net.minecraft.util.thread.BlockableEventLoop/FATAL]: Error executing task on Client
java.util.concurrent.CompletionException: java.lang.NullPointerException: Cannot invoke "io.netty.channel.Channel.eventLoop()" because "channel" is null
    at TRANSFORMER/sable@1.2.2/dev.ryanhcode.sable.network.packets.tcp.ClientboundSableUDPActivationPacket.handle(ClientboundSableUDPActivationPacket.java:51)
    ...
[Render thread/WARN] [net.minecraft.client.multiplayer.ClientCommonPacketListenerImpl/]: Client disconnected with reason: 无效的玩家数据
```

Then the player reconnects and it crashes again identically.

## Environment

- **Mod version**: sable-neoforge-1.21.1-1.2.2.jar
- **Minecraft**: 1.21.1 NeoForge
- **Launcher**: HMCL with Terracotta (EasyTier-based tunneling)
- **Relevant config**: `sable-common.toml`

## Root cause

[...]

Issue #1035, opened 2026-05-25
Server crash when not disabling udp pipeline


redstonek:
### Context

I was trying to install a specific modpack to a server and it kept crashing. After trying to narrow the issue I found out that Sable is responsible.
I believe it's related to my server provider not allowing too many udp ports, but what other mod could be using them?
Disabling the udp pipeline in config solves it, but it seems to be a half measure at best.

### Crash Report

https://mclo.gs/lrBkikr

Issue #852, opened 2026-05-10
Not able to join the server


Kisielelo320:
### Description

I cannot join my minecraft server but my friend can, I have all the same mods as him, I noticed that before I get disconnected it says in the logs that 
"[10May2026 19:51:41.094] [Netty Client IO #3/INFO][dev.ryanhcode.sable.Sable/]: Closed UDP channel!
[10May2026 19:51:41.094] [Netty Client IO #3/INFO][dev.ryanhcode.sable.Sable/]: Client UDP channel inactive"
How can I fix that?

### Game Log

https://mclo.gs/sGmB0RH

protimita:
Same issue, please escalate.
I have been playing on a server for weeks without issue until today. I was disconnected while playing with the same Close UDP channel log.

Dremzzz:
Same issue, my friends can join me but i can't join them. For some reason it happens only on 1.21.1 modpack we made. 🤔

Wepik5118:
Same problem, I hope the author will release a fix or help solve it, because no mods or changes in settings help

PawlosXpl:
It works for me after server restart but sable stops working after couple hours, and people can't join server. And in logs same issue as above

slava110:
My friend has the same problem, but he won't be able to access his PC for now.
Are you guys sure it's caused only by Sable? Have you tested this without any other mods, just with Sable (could be conflict)? Are you also on the latest version? Could you attach your own logs/debug logs maybe?

Sbgit-progaming:
do you have intel? if so probably duplicate of #1006

slava110:
Log posted here doesn't look like #1006 at all. And my friend is able to join singleplayer world without problems. The only problem is joining our server, in our case

Sbgit-progaming:
sorry, trying to hellp sort reports and didn't do my due diligence.

JacobWoodson:
Running into the same issue.
My pc can connect to dedicated server locally
My friends PC can connect to the server remotely
my partner's PC cannot connect to the server locally, gets UDP closed
my laptop cannot connect to the server locally or remotely, gets UDP closed.

Lonach:
[...]

Issue #850, opened 2026-05-10
Severe Desync from server when UDP is enabled


tmvkrpxl0:
### Description

For some reason sublevels are desynced severely for one of users of my modpack server. I disabled "attempt_udp_networking = false" and it is no longer desynced.

### Game Log

Unfortunately there was nothing of note worthy there. There was nothing printed from aeronautics or sable except occasional "received movement packet from non existent sublevel" which users that did not have desync issue also see.

PR #1141, opened 2026-06-08
Fix client hang during UDP bootstrap that prevents joining servers

JacobWoodson:
## Problem

`ConnectionMixin.sable$connect` (and its local-server twin) opens a UDP channel alongside the Minecraft TCP one when the player connects. The bootstrap currently ends with:

```java
final ChannelFuture channelFuture = new Bootstrap()...
        .connect(inetSocketAddress.getAddress(), inetSocketAddress.getPort());

channelFuture.syncUninterruptibly();
```

That `syncUninterruptibly()` runs on the `Server Pinger` (multiplayer list refresh) and `Server Connector` (actual join) threads. Those are the same threads that drive the TCP handshake / config phase of the Minecraft login. As long as the netty UDP `connect()` future hasn't resolved, no TCP login can make progress.

On most hosts the UDP `connect()` returns within milliseconds (UDP "connect" is just an OS-level associate-with-remote-address â€” no packets fly). But the netty call goes through `NioDatagramChannel.doConnect()` â†’ `javaChannel().connect(remoteAddress)`, which on Windows can stall when a third party is interposed on the socket layer: Windows Filtering Platform callouts from antivirus, "secure DNS" / enterprise proxies, some VPN clients, or layered service providers doing UDP inspection. On the affected client in this repro, that stall lasted 20-30+ seconds; the server's read-timeout fired first and closed the TCP socket. From the user's perspective, the login screen sat on "Connectingâ€¦" and then bounced with a generic "Disconnected".

This is the same shape of failure surfaced (with different proximate causes) in #852, #850 (workaround: `attempt_udp_networking=false`), #1035 (closed as dup of #482), and #1080 (NPE that this PR also fixes).

## Repro

* Sable NeoForge 1.21.1-1.2.2, Synthetic Horizons modpack, MC 1.21.1 + NeoForge 21.1.233.
* One Windows 11 laptop on the LAN couldn't join a dedicated server (locally to `192.168.1.90:25565` or remotely to `47.38.244.85:25565`). Other machines on the same LAN with the same pack joined fine.
* Pinger thread eventually succeeded in bringing up its UDP channel (server appeared in the list), but the Server Connector's UDP bootstrap was the long one â€” long enough to push the server past its config-phase timeout.
* With the patch in place on the same machine, UDP bootstrap completes in 1 ms on the same network path, and the player joins. The "drowning in the spawn ocean" verification was, uh, vivid.

## Alternatives considered

I went through a few options before landing on this one. Recording them so reviewers can push back if they prefer a different shape.

1. **Tell users to set `disable_udp_pipeline=true` and call it done.** This is the documented workaround today and was suggested on #1080. It works, but (a) it silently downgrades every user to TCP-only sub-level data forever rather than only on machines where UDP genuinely can't be set up, (b) the flag only short-circuits the server-side `ServerConnectionListenerMixin` â€” `ConnectionMixin.sable$connect` ignored it, so users following the doc were still hitting the bootstrap on the client, and (c) it doesn't address the root cause (a blocking call on the login thread).

2. **Move the UDP bootstrap off the login thread entirely** (run the `Bootstrap.connect(...)` on the netty event loop and `setUDPChannel` from a listener). Cleaner long-term shape -- UDP bring-up becomes purely async, no timing coupling with TCP login at all. I didn't take this because it changes the lifecycle assumptions made by `ClientboundSableUDPActivationPacket.handle` (which today reads `sable$getUDPChannel()` synchronously, expecting connect() to be done by the time the server's activation packet arrives) and by anyone reading the field elsewhere. That's a bigger change with a wider blast radius for a backport-friendly fix, so I deferred it. Happy to do it as a follow-up if the maintainers prefer.

3. **Bounded `await` with a timeout, fall back to TCP-only on failure.** What this PR does. The synchronous shape is preserved (so existing assumptions about `sable$getUDPChannel()` after `Connection.connect` returns still hold for the success path), but a slow or stuck `connect()` no longer holds the login thread hostage. The timeout (5 s) is generous enough that healthy networks always succeed (success path observed at <1 ms on every machine I tested) and short enough that the server-side login window (which is also ~20-30 s before the read-timeout fires) is preserved.

4. **Make the timeout configurable.** Considered; rejected for now. A config knob is overkill when the practical range is "<1 s succeeds, otherwise something is broken and we want to fail fast". If a future deployment surfaces a legitimate need for a longer wait, promoting `SABLE$UDP_CONNECT_TIMEOUT_MS` to `SableConfig` is one line.

## What this PR does

* **`ConnectionMixin.sable$connect`**: replace `syncUninterruptibly()` with `awaitUninterruptibly(SABLE$UDP_CONNECT_TIMEOUT_MS, MILLISECONDS)` (5 s). On timeout: cancel the future, install a listener that logs the late outcome (and closes the channel if it eventually succeeds after we already gave up), return â€” letting TCP login proceed. On `!isSuccess()`: log the cause, return. Wrap the bootstrap construction itself in try/catch so an unchecked exception during bootstrap can't break the connect call either.
* **`ConnectionMixin.sable$connectToLocalServer`**: same treatment for `LocalChannel`. Less critical (local channels rarely stall) but the inconsistency was hard to justify keeping.
* **Honor `SableConfig.DISABLE_UDP_PIPELINE` on the client**: short-circuit both connect injects when the flag is set. Previously only `ServerConnectionListenerMixin` checked it, which made the documented workaround a half-measure.
* **`ClientboundSableUDPActivationPacket.handle`**: null-check `connectionExtension.sable$getUDPChannel()` before dereferencing â€” fixes the NPE in #1080 (`Cannot invoke 'io.netty.channel.Channel.eventLoop()' because 'channel' is null`). When the channel is null, log a WARN and return; the client stays on TCP-only and the server's auth state for that connection just stays in `AWAITING_AUTH` until the connection closes. Also pattern-match the `connection.getRemoteAddress()` cast instead of an unchecked one.
* **Logging cleanup**: the existing `Starting remote client UDP channel future` line now carries the remote address and transport class so logs from affected users are immediately actionable. All failure paths log at WARN with the cause. Routine breadcrumbs (config-disabled short-circuit, late-success-after-cancel, success path with timing) are demoted to DEBUG so unaffected users don't see any new chatter.

### Success path
Byte-for-byte identical end-state -- the channel is set on the `ConnectionExtension`, `SableUDPChannelHandlerClient.channelActive` fires, `ClientboundSableUDPActivationPacket.handle` sees the channel and dispatches the auth response, server transitions `SableUDPAuthenticationState` to `AUTHENTICATED`. No timing changes when UDP comes up healthily (success observed at <1 ms across machines tested).

### Failure / timeout path (where this PR is doing the actual work)

The interesting question is "if UDP setup fails, what's broken for the player?" I traced the gameplay-affecting UDP call sites to answer that:

* `sendUDPPacket` has two production call sites in the codebase: `SubLevelTrackingSystem.java:367-397` (sub-level snapshot replication) and `SableCommand.java:58` (a debug echo command). The tracking system already has an **explicit, designed-in TCP fallback branch** keyed on `isConnectedTo(player)` â€” sub-level snapshots are sent as a `ClientboundBundlePacket(ClientboundSableSnapshotInfoDualPacket, ClientboundSableSnapshotDualPacket)` over TCP when UDP isn't authenticated. The packets are *Dual* (`SableUDPPacket, SableTCPPacket`) and the client decodes them identically off either transport; the only branch in `ClientSableInterpolationState.receiveSnapshot` on the receive mode is a debug-overlay string ("Networking through UDP/TCP"). The interpolation math, the snapshot buffer, and the rendered sub-level position are all identical.
* `isConnectedTo(player)` returns `false` cleanly in our failure path (`udpAuthStates` entry stays in `AWAITING_AUTH` because the client never replies with `SableUDPAuthenticationPacket`).
* `SableUDPServer.sendPings()` skips players that aren't `AUTHENTICATED`, so a fallback-to-TCP player generates no UDP keepalive traffic and is never bounced by the missed-pings logic.

So the practical effect of our timeout firing is: **the player joins, sub-levels work normally, the only observable difference is the F3 debug overlay reports "Networking through TCP"** (or "UNKNOWN" until the first snapshot arrives). The one real degradation is that under a **lossy** network, sub-level snapshots over TCP can show jitter from head-of-line blocking that UDP wouldn't have â€” but the failure mode that triggers this PR is at the OS socket layer, not on-wire packet loss, so affected users typically have clean networks and see no visible difference.

### Other behavioral notes

* **`disable_udp_pipeline` now means what it says** on the client â€” no UDP bootstrap, no UDP channel, no log lines. Previously the flag only short-circuited the server bootstrap, so a user setting it client-side (as #1080 recommends as a workaround) was still hitting the buggy bootstrap. This is a small behavior change for anyone relying on the side-effect that the client kept opening a UDP channel; I don't think such users exist, but worth calling out.
* **Stale auth-state entries**: in the timeout path, the server's `udpAuthStates` map keeps an `AWAITING_AUTH` entry for the connection whose client never replied. The map is a `WeakHashMap` keyed on `Connection` and entries are also pruned in `SableUDPServer.sendPings()` when `!connection.isConnected()`, so it cleans up on disconnect. No leak; called out for completeness.
* **No protocol change**, no packet format change, no jarjar change. Drop-in compatible across mixed 1.2.2 / patched-1.2.2 deployments â€” server-patched + client-unpatched (or vice versa) negotiates UDP exactly the same way it does today.
* **No new dependencies**. One `@Unique` field added: `SABLE$UDP_CONNECT_TIMEOUT_MS` (mixin-safe naming).

## AI assistance

This change was developed with the assistance of an AI coding assistant (Claude). I read every line, built the jar locally against `mc1.21.1-1.2.2-neoforge`, dropped it into the affected modpack instance, reproduced the original join failure on the affected machine, watched it succeed with the fix, and read the resulting logs to confirm the WARN/DEBUG split is what I want a future bug report to contain. I'm submitting this as my own contribution under the CONTRIBUTING.md terms.

## Related issues

* #1080 -- Client-side NPE in `ClientboundSableUDPActivationPacket` (fixed directly by the channel null guard)
* #852 -- Not able to join the server (consistent with this PR's repro for the subset of reports where the proximate cause is a slow OS-level UDP `connect()`)
* #850 -- Severe desync from server when UDP is enabled (different root cause; not addressed here)
* #1035 / #482 -- Server-side crash without disabling UDP pipeline (server-side; not addressed here)

### DIFF
--- a/common/src/main/java/dev/ryanhcode/sable/mixin/udp/ConnectionMixin.java
+++ b/common/src/main/java/dev/ryanhcode/sable/mixin/udp/ConnectionMixin.java
@@ -3,6 +3,7 @@ package dev.ryanhcode.sable.mixin.udp;
 import com.llamalad7.mixinextras.sugar.Local;
 import dev.ryanhcode.sable.Sable;
 import dev.ryanhcode.sable.SableClient;
+import dev.ryanhcode.sable.SableConfig;
 import dev.ryanhcode.sable.mixinterface.udp.ConnectionExtension;
 import dev.ryanhcode.sable.network.udp.SableUDPPacket;
 import dev.ryanhcode.sable.network.udp.handler.SableUDPChannelHandlerClient;
@@ -24,6 +25,7 @@ import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
 
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
+import java.util.concurrent.TimeUnit;
 
 @Mixin(Connection.class)
 public abstract class ConnectionMixin implements ConnectionExtension {
@@ -42,11 +44,14 @@ public abstract class ConnectionMixin implements ConnectionExtension {
         if (this.sable$udpChannel != null && this.sable$udpChannel.isOpen()) {
             this.sable$udpChannel = null;
 
+            Sable.LOGGER.debug("[sable-udp] closing UDP channel on disconnect, reason={}",
+                    disconnectionDetails != null ? disconnectionDetails.reason().getString() : "<no details>");
+
             channel.close().awaitUninterruptibly().addListener((x) -> {
                 if (x.isSuccess()) {
                     Sable.LOGGER.info("Closed UDP channel!");
                 } else {
-                    Sable.LOGGER.info("Failed to close UDP channel", x.cause());
+                    Sable.LOGGER.warn("Failed to close UDP channel", x.cause());
                 }
             });
         }
@@ -57,8 +62,17 @@ public abstract class ConnectionMixin implements ConnectionExtension {
         return this.sable$udpChannel;
     }
 
+    @Unique
+    private static final long SABLE$UDP_CONNECT_TIMEOUT_MS = 5_000L;
+
     @Inject(method = "connect", at = @At("TAIL"))
     private static void sable$connect(final InetSocketAddress inetSocketAddress, final boolean bl, final Connection connection, final CallbackInfoReturnable<ChannelFuture> cir) {
+        if (SableConfig.DISABLE_UDP_PIPELINE.get()) {
+            Sable.LOGGER.debug("[sable-udp] client UDP pipeline disabled via config; skipping bootstrap for {}", inetSocketAddress);
+            return;
+        }
+
+        final long startNs = System.nanoTime();
         final boolean useNativeTransport = SableClient.useNativeTransport();
 
         final Class<? extends Channel> channelClass;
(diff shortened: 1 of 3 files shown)


Milgurt:
I've been having this issue as well lately, with the server throwing this error 

> [19:15:57] [Netty Epoll Server IO #0/ERROR] [de.ry.sa.Sable/]: SableUDPPacket SNAPSHOT (ClientboundSableSnapshotDualPacket) was larger than expected, found 27 bytes extra

restarting server seems to temporarily solve issue, client and server are both using the same version of sable, 1.2.2

PR #1175, opened 2026-06-13
Fix renderPose caching

HoshinOFW:
Small logic error in ClientSubLevel stops renderPose from ever being cached. Fix is simple.


ryanhcode:
Thank you!

PR #1215, opened 2026-06-16
Add support for villager trading in sublevels

lefebvreb:
Add a few mixins to convert between global coordinates and sublevel coordinates so that villagers can correctly path to workstations (POIs), claim them, pick a trade, restock, etc.


Ocelot5836:
This will be redone soon and this isn't the correct way to handle trading

PR #1244, opened 2026-06-20
Add server config for physics steps

### LINKED ISSUE #1234: Suggestion: Automatically saves Sub-Level settings to the world
Make a the mod Sable can save my Sub-Level settings into my world, because every time I enter the world, it keeps resetting to 40

Ocelot5836:
This adds a server config that physics steps are written to. The runtime physics config is kept, but updated from the config value instead of not being saved anywhere. I used a server config so the data is saved per world instead of per instance.

Closes #1234


nizarghathbut:
Thanks a lot for adding this feature. Keep up the coding spirit😁

ryanhcode (approved):
Cool, awesome

PR #1274, opened 2026-06-25
Use the block pos when generating block collision.

### LINKED ISSUE #1273: Incompatibility with Immersive Engineering
Placing down an immersive engineering post with sable installed breaks due to this line:

https://github.com/ryanhcode/sable/blob/d3808e87acab2ee09f693bcc84e4c9133ed2fec1/sable_rapier/src/main/java/dev/ryanhcode/sable/physics/impl/rapier/collider/RapierVoxelColliderBakery.java#L75

IE posts depend on the having access to the block coordinates for their collision data, but since the function is called with zeroed out coordinates, the function fails with a null reference.

passing the block coordinates through, even if they're the shipyard coordinates, should be all that's needed to fix this.

RosaryMala:
Fixes https://github.com/ryanhcode/sable/issues/1273


ryanhcode:
It's intentional that only one voxel collider ever exists for a block-state, and that we don't pass in the block position of the blocks to generate that collider. Otherwise, the collider for every block of that block-state would be non-deterministic based on the block of that block-state that was placed first. The correct thing here is to make the IE post have a default collider or something of the like.

Issue #1328, opened 2026-07-04
[Crash] NullPointerException in RapierVoxelColliderBakery when processing Pixelmon MultiBlocks


MACTEP-ABATAP:
### Description
A server crash occurs when Sable's physics pipeline attempts to build physics data for a `MultiBlock` from the Pixelmon mod. 

When a Pixelmon multiblock updates or is placed, Sable triggers `RapierVoxelColliderBakery.buildPhysicsDataForBlock` which eventually calls `getCollisionShape`. Pixelmon's implementation of `getShape` for multiblocks attempts to query the `BlockGetter` for an adjacent blockstate, which returns `null`. This results in a `NullPointerException` when Pixelmon tries to invoke `.getBlock()` on that null state.

While the lack of null-safety is technically on Pixelmon's side, Sable is the one triggering the shape query (possibly during a state or context where Pixelmon doesn't expect it). Adding a safety check or a try-catch for third-party block shape queries during collider generation might prevent the server from crashing entirely.

### Environment
*   **Minecraft:** 1.21.1
*   **Modloader:** NeoForge
*   **Sable Version:** 2.0.3 (specifically `sable_rapier@1.21.1-2.0.3`)
*   **Pixelmon Version:** 9.3.16

### Steps to Reproduce
1. Install Sable (2.0.3) and Pixelmon (9.3.16) on a NeoForge 1.21.1 server.
2. Trigger a block state update involving a Pixelmon MultiBlock (e.g., placing or destroying a PC, Healer, or similar multiblock structure).
3. The server instantly crashes as Sable attempts to handle the block change.

### Crash Report
<details>
<summary>Click to expand relevant Stacktrace</summary>

```java
java.lang.NullPointerException: Cannot invoke "net.minecraft.world.level.block.state.BlockState.getBlock()" because the return value of "net.minecraft.world.level.BlockGetter.getBlockState(net.minecraft.core.BlockPos)" is null
	at TRANSFORMER/pixelmon@9.3.16/com.pixelmonmod.pixelmon.blocks.MultiBlock.getMultiBlockBoundingBox(MultiBlock.java:116)
	at TRANSFORMER/pixelmon@9.3.16/com.pixelmonmod.pixelmon.blocks.MultiBlock.getShape(MultiBlock.java:95)
	at TRANSFORMER/minecraft@1.21.1/net.minecraft.world.level.block.state.BlockBehaviour$BlockStateBase.getShape(BlockBehaviour.java:650)
[...]

PR #1329, opened 2026-07-04
Handle collision shape failures during collider baking

MACTEP-ABATAP:
### Summary

Makes Rapier collider baking more defensive when retrieving block collision shapes. Some blocks, especially from mods with complex or multiblock collision logic, may throw while Sable is building physics collider data. Without this, a single bad collision shape can crash collider baking and may leave the temporary block getter state configured for the failed block.

After this PR, collision shape retrieval failures fall back to an empty shape instead of crashing, and the temporary block getter state is always reset after retrieval.

### The problem

`RapierVoxelColliderBakery` temporarily configures `PhysicsColliderBlockGetter` with the block state being baked, then asks the block for its collision shape.

Previously, if either path threw:

  * `BlockSubLevelCollisionShape#getSubLevelCollisionShape`
  * `BlockState#getCollisionShape`

then `this.level.setup(Blocks.AIR.defaultBlockState())` would not run. That could leave the shared block getter in a stale state and crash the caller.

A mod returning `null` for a collision shape would also fail later at `shape.isEmpty()`.

### Changes

  * Wrap collision shape retrieval in defensive error handling.
  * Move `this.level.setup(childState)` into the guarded block.
  * Reset the temporary block getter state in a `finally` block.
  * Fall back to `Shapes.empty()` when shape retrieval fails.
  * Treat `null` collision shapes the same as empty collision shapes.
  * Keep the existing collider behavior unchanged for valid blocks and liquids.

### Testing

Not run in-game yet.

Recommended checks:

  * Start a world with the modpack that previously crashed during collider baking.
  * Verify the server no longer crashes when the problematic block is encountered.
  * Verify normal solid blocks, liquids, fences, and walls still produce expected Rapier collision data.
  * Check logs for the new warning output when a problematic block falls back to an empty collision shape.

### Notes for reviewers

This change intentionally prefers a safe empty collider over crashing. That means a problematic block may temporarily behave as non-solid for Rapier physics, but the game/server remains alive.

The `finally` block is the important part of the change: even when a mod throws while building a collision shape, `PhysicsColliderBlockGetter` is reset back to air before the method returns or continues.

### Known limitations

  * A block whose collision shape retrieval fails will be treated as empty collision data.
  * Because collider data is memoized per `BlockState`, a fallback result may stay cached for that bakery instance.
  * The current implementation logs to `System.err`; this can be replaced with the project logger later if preferred.

### DIFF
--- a/sable_rapier/src/main/java/dev/ryanhcode/sable/physics/impl/rapier/collider/RapierVoxelColliderBakery.java
+++ b/sable_rapier/src/main/java/dev/ryanhcode/sable/physics/impl/rapier/collider/RapierVoxelColliderBakery.java
@@ -1,5 +1,6 @@
 package dev.ryanhcode.sable.physics.impl.rapier.collider;
 
+import dev.ryanhcode.sable.Sable;
 import dev.ryanhcode.sable.api.block.BlockSubLevelCollisionShape;
 import dev.ryanhcode.sable.api.block.BlockWithSubLevelCollisionCallback;
 import dev.ryanhcode.sable.api.physics.callback.BlockSubLevelCollisionCallback;
@@ -13,6 +14,7 @@ import net.minecraft.core.BlockPos;
 import net.minecraft.world.level.BlockGetter;
 import net.minecraft.world.level.block.Blocks;
 import net.minecraft.world.level.block.state.BlockState;
+import net.minecraft.world.phys.shapes.Shapes;
 import net.minecraft.world.phys.shapes.VoxelShape;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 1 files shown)


MACTEP-ABATAP:
Potentially closes #1328

superintendent2521:
would it not be better to build a generic block collision hitbox so you still have collisons, depsite the fact that they might not be accurate to the model

MACTEP-ABATAP:
## Changes

- Wrap collision shape retrieval in defensive error handling.
- Move collision shape retrieval into `getCollisionShapeOrFallback`.
- Reset the temporary block getter state in a `finally` block.
- Fall back to `Shapes.block()` when shape retrieval fails.
- Treat `null` collision shapes as full block collision shapes.
- Use `Sable.LOGGER.warn` instead of writing directly to `System.err`.
- Keep the existing collider behavior unchanged for valid blocks and liquids.

## Testing

- Run `./gradlew :sable_rapier:compileJava --stacktrace`.
- Built test jars with `./gradlew :fabric:remapJar :neoforge:jar`.

Recommended in-game checks:

- Verify problematic blocks still have a generic full-block Rapier collider.
- Verify normal solid blocks, liquids, fences, and walls still produce expected Rapier collision data.
- Check logs for the new warning output when a problematic block falls back to a full block collision shape.

## Notes for reviewers

This change intentionally prefers a generic full-block collider over either crashing or treating the block as non-solid. The fallback may not perfectly match the block model, but it preserves collision behavior well enough for problematic modded blocks.

The `finally` block is still the important safety piece: even when a mod throws while building a collision shape, `PhysicsColliderBlockGetter` is reset back to air before the method returns or continues.

## Known limitations

- A block whose collision shape retrieval fails will be treated as full block collision data.
- Because collider data is memoized per `BlockState`, a fallback result may stay cached for that bakery instance.

PR #1393, opened 2026-07-25
Fixed Create mechanical rollers not filling correctly by bypassing block breaking distance check

adityapatn:
Create mechanical rollers' "Sloped Fill" and "Straight Fill" modes interact with blocks much further away than other actors, and so should be excluded from Sable's distance check on whether to cancel the interaction in testBreakingPosDist in BlockBreakingMovementBehaviourMixin.java. 

Error: Create Mechanical Rollers on Create trains do not fill below the first layer when Create and Sable are installed (Latest versions for Neoforge 1.21.1). 

Fix: Add check for whether the MovementBehaviour is a RollerMovementBehaviour in the distance check in testBreakingPosDist.


ryanhcode:
Should be fixed in main

PR #1419, opened 2026-07-31
Remove potentially problematic memory optimization in BlockSubLevelLiftProvider

KosmX:
Avoid using "globals to avoid allocation", this way JIT can do proper escape analysis, do proper optimizations (and make the function thread-safe too)

More details on escape analysis:
https://docs.oracle.com/javase/8/docs/technotes/guides/vm/performance-enhancements-7.html#escapeAnalysis

There are many reasons why I would not create globals to avoid allocation:
- Escape analysis can stack promote all the internal vectors (that cannot leave the function scope), stack promoted things can be promoted to registers
- even if allocated, allocations will be in current cache, it is less likely for the code to cache-miss.
- updating object fields would slow-down the function more (since GC barriers are a thing), this way that doesn't happen.
- since I can avoid TEMP, the math is a bit more readable.
- accidentally made the function thread-safe.

I was thinking on what might be worse with this, but I couldn't find many reasons. Some allocations will occur before JIT compiles the function, but that probably doesn't matter (since compilation will happen pretty fast)
Also because this patch removes a few fields, it might break some mods, so merging this might be a bit risky.

And since the function isn't a performance bottleneck anyway, I would say this PR is pretty low priority.

### DIFF
--- a/common/src/main/java/dev/ryanhcode/sable/api/block/BlockSubLevelLiftProvider.java
+++ b/common/src/main/java/dev/ryanhcode/sable/api/block/BlockSubLevelLiftProvider.java
@@ -23,26 +23,6 @@ public interface BlockSubLevelLiftProvider {
 
     Direction[] DIRECTIONS = Direction.values();
 
-    // memory optimization
-    Vector3d LIFT_FORCE = new Vector3d();
-    Vector3d LIFT_POS = new Vector3d();
-    Vector3d LIFT_NORMAL = new Vector3d();
-
-    Vector3d LIFT_VELO = new Vector3d();
-    Vector3d DRAG = new Vector3d();
-    Vector3d TEMP = new Vector3d();
-
-    /**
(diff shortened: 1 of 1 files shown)


ryanhcode:
have you done benchmarking to support this for this case, or is it simply speculative?

ryanhcode:
I haven't had great experiences with this sortof thing but I'm happy to be proven wrong. I typically see the vector allocation showing up heavily in profiling for CPU time, it's what sparked this pattern in the first place

KosmX:
It is speculative, I don't know when can I do benchmarking. or with what profiler.
But now I wanna know the truth. My hypothesis is that this function is simple, escape analysis won't be a problem, but it has to be tested.

What is your profiler setup? maybe I can use that.

KosmX:
Tomorrow I'll attempt to profile this with async-profiler, but I have no idea if this change is measurable.

PR #1421, opened 2026-08-01
Fix ticket map lookup key type mismatch and logging placeholder format

Spagles:
Fixed an issue in `SubLevelTicketLoadingSystem` where passing a `ServerSubLevel` instance instead of its `UUID` to `allTickets.get()` resulted in `null` lookups and prevented ticket pointers from persisting on unload. Also added a missing `{}` placeholder in `SubLevelSerializer` error logging.


ryanhcode:
Thank you!

though I am curious, what scenario you were in where a force-loaded sub-level was getting unloaded lmao?

pootism-1988:
I don't code but you should look into this
Screenshot_20260803_125603_Reddit

PR #1432, opened 2026-08-03
fix(rapier): don't panic on JNI calls referencing removed physics bodies

Bazzirovan:
## What happened

I run a dedicated NeoForge 1.21.1 server (Windows) with Sable, and my server kept dying during normal play. Two crash signatures, same underlying cause:

- `ServerHangWatchdog` reports where a tick took 60000+ seconds, with the stuck frame in `Rapier3D.step(Native Method)`
- `NullPointerException` on the Java side (e.g. `RopePhysicsObject.setFirstSegmentLength`) that followed a native abort

## Root cause

The Rust JNI bridge panics when a call references a physics object that is already gone: `rigid_bodies[&id]`, `.expect("No rigid body for id")`, `.expect("No kinematic contraption with given ID!")`, and a few `.unwrap()`s on removed collider info.

A panic in Rust that crosses the JNI boundary doesn't unwind cleanly — it kills the whole JVM. On my server this triggered whenever a sublevel unloaded (or unloaded and re-created quickly) while a JNI call still referenced a stale ID. The most common trigger in practice was rope endpoints: a rope anchored to a sublevel body whose collider had already been removed from the scene.

## Fix

Replaced the panicking lookups with Option guards, keeping the existing behavior for everything that resolves correctly:

- `get_rigid_body` / `get_rigid_body_mut`: return `Option` instead of `expect()`
- `getPose`: fall back to the identity pose if the body is missing
- `removeSubLevel` / `removeBox` / `removeKinematicContraption`: no-op when the entry is missing
- `setMassProperties`, `teleportObject`, `wakeUpObject`, velocity/force applicators: no-op when the body is missing
- `createKinematicContraption`: drop the stale mount reference instead of panicking
- `add*Constraint`: return an invalid handle (`-1`) when an endpoint is missing. On the Java side `isConstraintValid(-1)` resolves to `false`, `assertValid()` guards mutation, and the pipeline turns it into a failed constraint creation rather than a crash. The sentinel was chosen as `-1` (not `0`) because the first valid impulse joint naturally receives handle 0 (index 0, generation 0), making 0 ambiguous as an error value; `-1` is impossible as a native handle
- `rope::tick`: drop rope attachments whose endpoint collider is no longer registered, same as the existing handling for joints that vanish
- rope JNI entry points (`queryRope`, `removeRope`, `setRopeFirstSegmentLength`, `removeRopePointAtStart`, `addRopePointAtStart`, `wakeUpRope`, `setRopeAttachment`): no-op when the rope id is stale or the attachment's sub-level body is gone, instead of unwrapping and aborting the JVM
- `setCenterOfMass`, `setLocalBounds`, `addChunk` (sub-level collider upload): no-op when the level collider with that id has already been removed from the scene (race: `SubLevelPhysicsSystem.onStatsChanged` ticking after the collider was unloaded)

The `-1` ground-body special case in the constraint functions is untouched.

## Verification

Rebuilt the Windows natives with these changes and ran the result on the live server. Before the patch: repeated `RopeHandle` NPE crashes and watchdog freezes. After: several days of uptime with ropes, sublevel ships and constraints in regular use, no crashes. I also built on the Rust side to make sure nothing else breaks (`cargo build` clean).

Scope note: I only touched the paths that actually crashed on my server — the dispatcher/contact-event code that indexes colliders during a step is left alone since those colliders are still alive while the step runs.

Files changed:
- `sable_rapier/src/main/rust/rapier/src/lib.rs`
- `sable_rapier/src/main/rust/rapier/src/contraptions.rs`
- `sable_rapier/src/main/rust/rapier/src/boxes.rs`
- `sable_rapier/src/main/rust/rapier/src/joints.rs`
- `sable_rapier/src/main/rust/rapier/src/rope.rs`

### DIFF
--- a/sable_rapier/src/main/rust/rapier/src/boxes.rs
+++ b/sable_rapier/src/main/rust/rapier/src/boxes.rs
@@ -81,7 +81,10 @@ pub extern "system" fn Java_dev_ryanhcode_sable_physics_impl_rapier_Rapier3D_rem
         let sim_data = &mut *sim_data;
         let mut sable_data = scene.sable_data.write().unwrap();
 
-        let handle = sable_data.rigid_bodies[&(id as LevelColliderID)];
+        let Some(handle) = sable_data.rigid_bodies.get(&(id as LevelColliderID)) else {
+            return;
+        };
+        let handle = *handle;
         sim_data.rigid_body_set.remove(
             handle,
             &mut sim_data.island_manager,
--- a/sable_rapier/src/main/rust/rapier/src/contraptions.rs
+++ b/sable_rapier/src/main/rust/rapier/src/contraptions.rs
@@ -36,11 +36,8 @@ macro_rules! extract_jint_array {
 fn get_kinematic_collider_info(
     sable: &mut SableSceneData,
     id: jint,
-) -> &mut ActiveLevelColliderInfo {
-    sable
-        .level_colliders
-        .get_mut(&(id as LevelColliderID))
-        .expect("No kinematic contraption with given ID!")
+) -> Option<&mut ActiveLevelColliderInfo> {
+    sable.level_colliders.get_mut(&(id as LevelColliderID))
 }
 
 #[unsafe(no_mangle)]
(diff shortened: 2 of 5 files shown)


ryanhcode:
I'm not a fan of lots of stuff here, like falling back to the identity pose if the body is missing, or just blindly covering up any places where we're still referencing removed bodies. Also not fond of AI generated contributions.

What is the actual crash you were encountering? I'd like to fix the callsite that is keeping hold of removed bodies rather than covering this up in the natives.

PR #1444, opened 2026-08-07
Removed LittleTiles as incompatibility

CreativeMD:
Hey there,

I am the author of LittleTiles and worked on supporting sable the past weeks. It is finally finished and out. Would be great if the incompatibility marker is removed. Version 1.6.0pre223 and newer support sable.

In Regards
CreativeMD


ryanhcode:
Thank you! It'll be marked as compatible in the next patch

CreativeMD:
Thanks a lot. Also got a question if there is a weight system? If yes is there a way for me to support it so I can calculate the weight based on the volume of the tiles inside my blocks?

Issue #1297, opened 2026-06-29
Very specific bug: Elevator pulley on sub level with 58 or more input RPM Tp players very high.


netgoatfr:
### Description

So, when using the elevator made using a elevator pulley, with 58 or more rpm, it will teleport players pretty high in the sky above the elevator.

No mod conflict, sable / create / aeronautics are all up to date.
It also happens sometime when the elevator is still going up (see video)

### Game Log

Nothing is being logged when the bug happens.

netgoatfr:
As issues take time to be resolved, i figured i would host myself the video
here it is https://averis.cc/bug.mp4

netgoatfr:
it appear simply touching the elevator while it's moving will trigger the tp, which is about a 50 blocks Y coordinate increase

PR #1446, opened 2026-08-07
Fix Create elevator clipping on Sable sublevels

### LINKED ISSUE #1304: Standing on a moving Create elevator inside a Sable sub-level desyncs the player
## Environment
- Sable 2.0.3 (NeoForge, MC 1.21.1)
- Create 6.0.10
- Create: Aeronautics 1.3.0
- Sodium 0.8.x + Iris rendering stack (as required by Sable 2.0.x)

## Summary
When a player **stands** (not seated) on a Create **Elevator Pulley** contraption that has been assembled **inside a Sable sub-level** (e.g. an airship the elevator shaft is glued into), riding the elevator desyncs the player:

- The player is **left behind** on the previous floor while the cabin moves on without them.
- The client console spams **`<player> moved too quickly!`** with a phantom vertical delta of ~280 blocks (X/Z ≈ 0).
- The ship and cabin **flicker / briefly disappear** during the movement (the camera lurches into plot space).

Riding the same elevator **seated** (as a passenger in a glued seat) works perfectly — smooth transition, no flicker, no desync. Rotating contraptions on the sub-level (e.g. bearings/propellers) are also unaffected.

## Diagnosis
The elevator contraption itself is captured into the sub-level correctly — it initializes at plot-space coordinates with `inSubLevel = true`, identical to other contraptions that work. The breakage is specific to the **standing-rider carry**.

`ContraptionCollider.saveClientPlayerFromClipping` / `savePlayerFromClipping` (the client-side anti-clip safety) reposition a standing rider using:

```java
double yStart = entity.maxUpStep() + contraptionEntity.getY() + yStartOffset;
...
entity.setPos(entity.getX(), yStart - shortestDistance, entity.getZ());
```

`contraptionEntity.getY()` is the contraption entity's **raw** position, which for a sub-level contraption is **plot-space** (far-away coordinates), not the rendered world-space position. `ContraptionColliderMixin` transforms the values used inside `collideEntities` into the sub-level frame, but **`saveClientPlayerFromClipping` / `savePlayerFromClipping` are not covered**, so a standing rider is `setPos`'d toward plot space — producing the ~280-block phantom Y delta (→ "moved too quickly" → rubber-band / left behind) and the camera lurch (→ render flicker).

This lines up exactly with the observed behavior:
- The safety only arms for `TranslatingContraption` (`ElevatorContraption extends PulleyContraption extends TranslatingContraption`), so rotating contraptions (bearings) never reach it.
- It returns early for passengers (`isPassenger()`), so a **seated** rider is unaffected.
- Only a **standing** rider on a **translating** contraption inside a sub-level triggers it.

## Steps to reproduce
1. Build a Sable airship sub-level and assemble it so it flies.
2. Glue a Create Elevator Pulley + shaft into the ship (so it becomes part of the sub-level) with at least two floors / elevator contacts.
3. Assemble the elevator.
4. **Stand** on the cabin and send it to another floor.
   - Expected: ride smoothly to the floor.
   - Actual: left behind on the old floor; client logs `moved too quickly!` (~280 Y delta); ship/cabin flicker.
5. Repeat **seated** in a glued seat → works correctly.

## Suggested fix
Transform the contraption Y used by `saveClientPlayerFromClipping` / `savePlayerFromClipping` into the sub-level frame (the same way the `collideEntities` inputs are already redirected), or skip these anti-clip safeties entirely for contraptions inside a sub-level (the sub-level-transformed `collideEntities` carry already handles riders correctly).

## Workarounds
- Ride the elevator **seated** (the passenger path is unaffected).
- Skipping `savePlayerFromClipping` when `getContaining(contraptionEntity) != null` (i.e. the contraption is inside a sub-level) restores correct standing-rider behavior in testing, with no observed downside for normal-speed elevators.

## Elevator Bug (Without Fix Applied)
https://github.com/user-attachments/assets/edbdb656-23a8-4d4a-8587-9b4cc8e717f6

## Elevator Bug (With Fix Applied)
https://github.com/user-attachments/assets/4d300b02-a6d8-4213-a9a3-47bc00704241

VictorioMo:
This PR fixes Create elevator player clipping/teleportation on Sable sublevels.

The issue appears to come from ContraptionCollider.savePlayerFromClipping() mixing global player coordinates with sublevel/contraption coordinates. This could cause the player's Y position to be set roughly 100–200 blocks above the elevator.

The fix converts the relevant contraption Y and anti-clipping raytrace coordinates between Sable sublevel space and global space.

Fixes #1304
Related to #673 and #1297

Test video: https://youtu.be/cQeV1smUpY0

I tested the fix after reloading the world and after assembling a new elevator, and could no longer reproduce the teleport. Other scenarios have not yet been tested.

### DIFF
--- a/neoforge/src/main/java/dev/ryanhcode/sable/neoforge/mixin/compatibility/create/contraptions/ContraptionColliderMixin.java
+++ b/neoforge/src/main/java/dev/ryanhcode/sable/neoforge/mixin/compatibility/create/contraptions/ContraptionColliderMixin.java
@@ -17,6 +17,9 @@ import org.spongepowered.asm.mixin.Unique;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Redirect;
 
+import com.simibubi.create.content.contraptions.ContraptionHandlerClient;
+import net.minecraft.world.phys.BlockHitResult;
+
 @Mixin(value = ContraptionCollider.class, remap = false)
 public class ContraptionColliderMixin {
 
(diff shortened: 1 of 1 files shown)


xYutake:
I ran into this same bug and fixed it independently before finding this PR, so here's my data point.

I used a single redirect on toGlobalVector in savePlayerFromClipping, converting the hit back to world space via logicalPose().transformPosition. That resolved it on Create 6.0.10, Simulated 1.3.2, Sable 2.0.5. Same approach as your sable$toActualGlobalVector.

I haven't run your version, so I can't speak to the other two redirects. Two questions in case it's useful:

Was the getY() redirect needed in your testing? From tracing it, the plot offset appeared to cancel out already, since the callers subtract the same value that savePlayerFromClipping adds back.

And for rayTraceContraption, I left that path alone because ContraptionHandlerClientMixin seemed to already handle the conversion. Possible there's a case where it doesn't that I didn't hit.

Issue #1269, opened 2026-06-24
Crash on world load: BEX64 in sable_rapier_x86_64_windows.dll (Rapier3D.initialize) - Sable 2.0.3 + Create Aeronautics 1.3.0 / NeoForge 1.21.1


KirillKremtsev:
### Context

**What I was doing:**  
I created a new world with Sable 2.0.3 and Create Aeronautics 1.3.0 on NeoForge 1.21.1. The game crashes immediately when the world loads, specifically during the initialization of the Rapier physics pipeline (`Rapier3D.initialize`). The crash is reproducible 100% of the time. Sometimes, after a crash and cleaning the `.sable` folder, the world loads once, but on subsequent re-entry, it crashes again.

**What I've already tried:**  
- Changing Java runtime (Microsoft OpenJDK → Eclipse Temurin 21.0.11)
- Reducing/increasing memory (-Xmx8192M, -Xmx10000M)
- Disabling sound (all volume sliders to 0)
- Disabling AVX512 (via BIOS)
- Removing all other mods (only Sable + Create Aeronautics left)
- Deleting the `.sable` folder and letting it regenerate
- Adding the `.sable` folder and `sable_rapier_x86_64_windows.dll` to Windows Defender exclusions
- Checking for hanging `javaw.exe` processes in Task Manager
- Trying older Sable versions (1.1.3, 1.2.2) – same crash

**System information:**  
- OS: Windows 11 Pro (24H2, Build 26100.8521)  
- CPU: AMD Ryzen 7 7800X3D  
- GPU: NVIDIA GeForce RTX 5070 Ti (driver 610.62)  
- RAM: 32 GB  
- Java: Eclipse Adoptium Temurin 21.0.11+10 (64-bit)  
- Mod loader: NeoForge 21.1.233  
- Sable version: 2.0.3 (also tested 1.2.2 and 1.1.3)  
- Create Aeronautics: 1.3.0  
- Other mods: Create 6.0.10, Flywheel 1.0.6, Sodium 0.8.12 (and dependencies) – but crash also occurs without Sodium.

**Crash details:**  
- Java error: `EXCEPTION_UNCAUGHT_CXX_EXCEPTION (0xe06d7363)` in `sable_rapier_x86_64_windows.dll` at `Rapier3D.initialize`  
- Windows Event Log: `BEX64` (0xc0000409) in `OpenAL.dll` (offset 0xa2b05) – but this seems collateral, as the crash originates in Rapier.

[...]

PR #1448, opened 2026-08-07
Fix Rapier initialize hard-crash when launcher stdout pipe closes

JoaoVitorBaggio:
## Summary
- Closing CurseForge/Overwolf while Minecraft stays open breaks the process stdout pipe. `fern` then panics on `info!` during `Rapier3D.initialize` (Windows os error 232 / "pipe is being closed"), even though the physics scene was already constructed.
- That panic previously crossed the JNI boundary and hard-killed the JVM (`EXCEPTION_UNCAUGHT_CXX_EXCEPTION` / `hs_err`) with no Minecraft crash report.
- Wrap Rapier native stdout logging in an `IgnoreClosedPipe` writer, and `catch_unwind` `initialize` the same way `tick`/`step` already do, so failures become Java exceptions / crash reports instead.

Related to #1269 (same `Rapier3D.initialize` / `hs_err` symptom; this PR definitively fixes the launcher-closed-stdout repro; other #1269 causes may still need separate investigation).

## Test plan
- [x] Local NeoForge 1.21.1 / Sable 2.0.3 client: join world with CurseForge open — OK
- [x] Leave world, close CurseForge/Overwolf, rejoin same world — previously `hs_err` / panic; with this native build — world loads
- [x] Intermediate build with only `catch_unwind` produced a Minecraft crash report naming the closed-pipe panic (confirmed root cause)
- [ ] CI rebuilds multi-platform `sable_rapier` natives so the Rust fix ships in a release jar


ryanhcode:
Fixed in main

PR #1483, opened 2026-08-18
fix: skip massless blocks in sable$buildProperties

### LINKED ISSUE #457: Server Crash (NPE) for weightless contraptions
### Context

Image

### **Description**
A `java.lang.NullPointerException` occurs when interacting with a Create contraption (specifically by placing a torch or updating blocks) while using the **Sable** physics mod. The crash is triggered during the entity ticking process.

### **Technical Details**
* **Minecraft Version:** 1.21.1 
* **NeoForge Version:** 21.1.227 
* **Sable Version:** 1.1.3 
* **Create Version:** 6.0.10 
* **Error Type:** `java.lang.NullPointerException` 
* **Error Message:** `Cannot invoke "org.joml.Vector3dc.negate(org.joml.Vector3d)" because the return value of "dev.ryanhcode.sable.api.physics.mass.MassTracker.getCenterOfMass()" is null` 

### **Steps to Reproduce**
1.  Set up a **Create Train Station** and assemble a train/contraption.
2.  Place a **Torch** (or any block that updates the contraption structure) on the assembled contraption.
3.  The server immediately crashes when the `AbstractContraptionEntity` attempts to tick and update its physical properties via Sable's Mixin.

### **Analysis**
The crash occurs in `com.simibubi.create.content.contraptions.AbstractContraptionEntity.sable$buildProperties`. Sable's mass tracking system fails to return a valid center of mass (`getCenterOfMass()` returns `null`), causing the physics calculation to fail when it attempts to negate that vector.

### **Environment**
* **OS:** Linux (amd64) version 6.1.0-44-amd64 
* **Java:** 21.0.10 (Eclipse Adoptium) 
* **Affected Entity:** `minecraft:minecart` (Entity ID: 7) carrying an `OrientedContraptionEntity` 

### Crash Report

https://mclo.gs/0RXjAHK

awakaxis:
This pr adds a check for massless blocks in buildProperties, skipping them alongside air blocks. This way, buildProperties for contraptions that consist of only massless / non solid blocks will return early, skipping the unchecked getCenterOfMass().negate() call that causes the crash.

I'm not sure if sable intends for massless / non solid blocks to potentially be a BlockSubLevelLiftProvider or floating material. If so, buildProperties will no longer set up these properties for massless blocks and some other fix will need to be found.

fix #457

### DIFF
--- a/neoforge/src/main/java/dev/ryanhcode/sable/neoforge/mixin/compatibility/create/contraptions/AbstractContraptionEntityMixin.java
+++ b/neoforge/src/main/java/dev/ryanhcode/sable/neoforge/mixin/compatibility/create/contraptions/AbstractContraptionEntityMixin.java
@@ -153,7 +153,7 @@ public abstract class AbstractContraptionEntityMixin extends Entity implements K
             final StructureTemplate.StructureBlockInfo info = entry.getValue();
             final BlockState state = info.state();
 
-            if (state.isAir()) continue;
+            if (state.isAir() || PhysicsBlockPropertyHelper.getMass(this.sable$blockGetter(), blockPos, state) == 0) continue;
 
             if (this.sable$localBounds == null) {
                 this.sable$localBounds = new BoundingBox3i(blockPos.getX(), blockPos.getY(), blockPos.getZ(), blockPos.getX(), blockPos.getY(), blockPos.getZ());


ryanhcode:
Already fixed in 2.0.5

PR #1531, opened 2026-09-02
Update build instructions following recent change

boisei0:
A recent commit (effec5f3da06556b1ca16a288499d144027940be) separated `sable_rapier` into its own module. The build instructions on the README hadn't been updated to reflect that change. This tiny patch does that.


ryanhcode:
Thank you!

PR #1576, opened 2026-09-19
Keep Quark's glass item frames aboard sub-levels

Gerry3010:
Quark's glass item frame (`quark:glass_frame`) is missing from `sable:wall_entities`, so `EntitySubLevelUtil.shouldKick` returns `true` for it. The frame is therefore projected out of the plot as soon as it is placed — `ServerLevel.addFreshEntity` → `EntitySubLevelUtil.kickEntity` on the server, `Entity.recreateFromPacket` on the client. Out there it holds parent-level coordinates instead of sitting on the sub-level's block grid, so it no longer lines up with the ship or follows it, and with no supporting block behind it `ItemFrame.survives()` fails on the next check and the frame drops.

`org.violetmoon.quark.content.building.entity.GlassItemFrame extends ItemFrame`, so it is the same kind of entity as `minecraft:item_frame` and `minecraft:glow_item_frame`, which are listed already — as are the frames from Exposure, Labels and Immersive Paintings. The glowing variant shares that entity type (it differs only by a synched `IS_SHINY` flag), so the single entry covers both items.

`required: false`, so the tag stays inert without Quark installed.

The symptom was observed on NeoForge 1.21.1 with Sable 2.0.5 and Quark 4.1-482, and the cause traced to `shouldKick`. In-game confirmation of the patched tag is still pending on our server; the change is data-only and flips exactly that check.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Gerry3010:
Confirmed in game since opening this: with `quark:glass_frame` in the tag, glass item frames placed on a sub-level now stay on the ship's block grid and no longer drop. Tested on a NeoForge 1.21.1 server running Sable 2.0.5, Create Aeronautics 1.3.1 and Quark 4.1-482, on an assembled airship in motion.

Before the change the frame was projected out of the plot the moment it was placed and `ItemFrame.survives()` dropped it on the next check, exactly as described above.

GaMiR9195:
2 lines change? Good.

